PR #18, opened 2019-02-25
removing wrong copy of file.

TimMangan:
Two copies of the Readme are present in the PsfLauncher folder. This removes the "wrong" one.


dhoehna:
Good job on catching that.

PR #19, opened 2019-02-25
Launch target argument and launching files via shell

TimMangan:
This pull request includes the following:

- Allow for specification of command line arguments for the target executable application in the json file. This is a requirement since MSIX shortcuts do not have command line arguments (at present).

- Allow the target "executable" to be a file instead of an exe and use the shell to launch it.  This is will support conversion of commonly used shortcuts to documents, HTAs, and URLs without having to hard code the app to handle the document in the package (this should be resolved by the FTA default setup for the user on their system).

- Small improvements to the debug logging of PsfLaunch/Runtime to make the output easier to understand.


PsfLauncher/main.cpp
```diff
@@ -125,53 +122,109 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
+        memset(&shex, 0, sizeof(SHELLEXECUTEINFO));
+        shex.cbSize = sizeof(shex);
+        shex.fMask = SEE_MASK_NOCLOSEPROCESS;
+        shex.hwnd = (HWND)NULL;
+        shex.lpVerb = NULL;  // just use the default action for the fta
+        shex.lpFile = nonexePath.c_str();
+        shex.lpParameters = exeArgString;
+        shex.lpDirectory = dirStr ? (packageRoot / dirStr).c_str() : nullptr;
+        shex.nShow = static_cast<WORD>(cmdShow);
+        //shex.hInstApp = 0; // for output
```
dhoehna: Did you leave these commented lines in here for a reason?  I see you say "for output".  Should these lines be un-commented for a specific reason?

PsfLauncher/main.cpp
```diff
@@ -125,53 +122,109 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
-        return ERROR_INVALID_HANDLE;
-    }
+        default:
+            return ERROR_INVALID_HANDLE;
+        }
 
-    DWORD exitCode;
-    if (!::GetExitCodeProcess(processInfo.hProcess, &exitCode))
+        DWORD exitCode;
+        if (!::GetExitCodeProcess(processInfo.hProcess, &exitCode))
```
dhoehna: Thank you for making sure there are brackets around the single-line if.

PsfRuntime/Config.cpp
```diff
@@ -402,34 +403,40 @@ PSFAPI const psf::json_value* __stdcall PSFQueryConfigRoot() noexcept
...
+	{
+            LogCountedStringW("Compare against json id", appId.data(), appId.length());
+        }
+	if (iwstring_view(appId.data(), appId.length()) == applicationId)
         {
             return &appObj;
         }
     }
-	Log("No Matches");
+    if (verbose
```
dhoehna: Forgot the closing ) on this if statement.

PsfLauncher/main.cpp
TimMangan: No. Lines were replaced by memset.  Removed commented lines.

PR #37, opened 2019-04-17
Support for configurable file redirection

TimMangan:
The changes in this PR are to enable the enterprise customer to control the redirection done by the Psf FileRedirectionFixup.  It adds the following features:

1. Expansion of the base/pattern specification pair to add:
- A redirection base folder (replacement target for the redirection to localappdata).
- An exclusion option (used to disable Copy-on-Access for executables and others. Allows the redirection array to specify exclusion patterns in one element, and a wildcard pattern in the next).
- A “read-only” option (used to lock-down a copy of a configuration file by marking the copy read-only).

2. Added AppData/Local, AppData/Roaming, and System to the KnownID list.
3. Added a few more common folders seen in App-V packages converted to MSIX using the MPT
4. Added special code to perform the missing Copy-on-Access from the package VFS AppData/Roaming & Local & other locations.

With these changes, it is possible to run packages that contain settings and data in all of the file areas that older applications might put them, to allow updates to those files as appropriate, and to control the redirection to send the changed files to a home drive to support roaming scenarios such as VDI and RDS, and for desktop replacement strategies without a loss of settings and data.


dhoehna:
I tested your code on my dev box and on an RS5 VM.  There is a bug and an issue I would like to address.
1. There is a bug on PathRedirection.cpp line 516.  This will fail if g_packageRootPath is longer than deVirtualizedPath.full_path.  No exception is thrown if this happens because the call is marked with extern "C" and should not leak any exceptions.
2. The WritablePackageRoot logic needs to stay in PSF.  This is to prevent apps from breaking when they write to the package root path.

You change might still adhere to the writable package root path logic.  But I wasn't able to test your changes because of the bug.

I can give you the package I am using to test with if you want.

TimMangan:
I’ll take care of #1.  I noticed additional issues with a strange app I’m testing and plan to improve the code in general, so please hold off additional testing for now.

The additional work includes fixes for new test cases I added over the weekend, and covers issues with checking for file existence (GetAttributes), and handling path requests with URI characters needing escaping, and a case of a file reference of “file:\c:\…”.

As for #2.  I believe that I handle the issue differently. 😉    There are a lot of locations the app might try to change things in, and the root folder is just one example. Writes to files in the root (or sub root other than VFS) are redirected, it’s just not under a new special folder.

Regards,

Tim

TimMangan:
On #1, the new code I’m working with always calls RedirectedPath using a path based on the g_packageRootPath.  The idea is that you pass in the path using the package form, and this function always returns a path that aligns with where you would redirect to if anything is redirected.

But in case someone calls it differently in the future I’m putting in protection around it to just append the input path instead.

I’d have sent a heads up on the additional work, but didn’t see it assigned yet so I was waiting until I finish up these changes.

Regards,

Tim

dhoehna:
Hi Tim,

Thank you for your quick response.

About item number 1.  I was testing PSF with an underlying C# application.  I got an error that the .NET Framework v4.0 did not exist.  Just letting you know.  I'm guessing the error came from incorrect redirection.

Also, good job on adding more tests.  Tests are good.

On number 2.  I realized I didn't explain fully why the writable package root is in place.  According to your response you redirect writes to the package root as well, just to a different location.  The old version of PathRedirection.cpp did that same thing, I just added a spcial case called the WriteablePackageRoot.

The reason why the path gets redirected differently is to prevent apps breaking on an upgrade when writes happen under the writable package root.

Example.
I write to C:\Program Files\WindowsApp\PackageFullName\MyFile.txt.
Under the old logic, this path would be redirected to
%LOCAL_APP_DATA%\VFS\C$\Program Files\WindowsApps\PackageFullName\MyFile.txt.

With the new logic, the redirected path is
%LOCAL_APP_DATA%\Packages\PackageFamilyName\LocalCache\Local\Microsoft\WritablePackageRoot\MyFile.txt

Since the redirected path in the new logic does not contain the Package Full Name an upgraded version will be able to read from the same location as the previous version.

This is why the writable package root exists, to make sure upgraded apps who write to their install location don't break.

This is why the additional logic for the WritablePackageRoot exists.  PSF needs to keep that.

TimMangan:
#1 I located another call today into the function from FindFirstFile and am addressing it also.  But as noted, I added protection in the function anyway.

#2 Under the new-new logic it is now redirected to:
%LOCAL_APP_DATA%\Packages\PackageFamilyName\LocalCache\MyFile.txt

So LocalCache will have everything relative to the package structure, but now under the PackageFamilyName.

…which brings up the question on why PackageFamilyName is used (which include the hash) instead of PackageName?  If the new version uses a different certificate the family name changes.

Today I found app making file references to “file:\c:” and another to “blob:”, so these too will be handled.  I will continue testing this week with a goal of completing that testing and updating the PR before I leave Saturday (for Switzerland for 2 weeks).

Regards,

Tim

TimMangan:
On #2 The previous email was on memory. This email is after testing a similar case to yours without specifying an alternate redirection location…

If you specify a home drive, it works like I said under the home drive.  If you don’t specify an alternate location, and use default the redirection location, the problem is still solved but because my redirection gets re-redirected by the MSIX runtime, the path actually looks like:
%LOCAL_APP_DATA%\Packages\PackageFamilyName\LocalCache\Local\VFS\PackageCache\PackageFamilyName\MyFile.txt

Which works, but is butt ugly. Maybe we can do better!

Tim
From: TIMOTHY MANGAN
Sent: Tuesday, April 23, 2019 5:08 PM
To: Microsoft/MSIX-PackageSupportFramework <reply@reply.github.com>
Subject: RE: [Microsoft/MSIX-PackageSupportFramework] Support for configurable file redirection (#37)

#1 I located another call today into the function from FindFirstFile and am addressing it also.  But as noted, I added protection in the function anyway.

#2 Under the new-new logic it is now redirected to:
%LOCAL_APP_DATA%\Packages\PackageFamilyName\LocalCache\MyFile.txt

So LocalCache will have everything relative to the package structure, but now under the PackageFamilyName.

…which brings up the question on why PackageFamilyName is used (which include the hash) instead of PackageName?  If the new version uses a different certificate the family name changes.

Today I found app making file references to “file:\c:” and another to “blob:”, so these too will be handled.  I will continue testing this week with a goal of completing that testing and updating the PR before I leave Saturday (for Switzerland for 2 weeks).

Regards,

Tim

dhoehna:
Hey Tim,

We are using Package Family Name because that is what Microsoft Games uses.  In fact, this writable package root is now used by games in Microsoft and we can't break backwards compatibility.

I spoke with someone in games about the redirection for writing into the install package.  PSF needs to be able to redirect all writes and reads to the WritablePackageRoot.  The WritablePackageRoot does need to stay in PSF going forward.

TimMangan:
A solution that will likely be a problem down the road, made for less than stellar reasons to stay compatible with something we don’t need to be compatible with; I’ll put it back.

Probably won’t happen this week as I’d have to start the testing all over.  I’ll update GitHub with what I have by the end of the week, but hold off processing the PR until I’m back and can restore this gem.

Tim

dhoehna:
Hi Tim,

Thank you.  It does help the games team greatly that the WritablePackageRoot logic stays in PSF.  I spoke to the games team to see if we could redirect to a different location, just so long as an app that is upgraded is redirected to the same location.  They require that any writes to the install directory be to the location in PSF master.

How would this be a problem down the road?  I'm not stating that you are being rude, I'm just curious.  I can talk to my team and we can figure out a solution that would make both worlds happy.

TimMangan:
Don’t mind me.  Just venting at inconsistencies.

But I should mention that the inclusion of the cert hash in the package family name means app upgrades loosing settings and data if the cert changes.  Certs expire and people get new ones.  Companies get acquired and get new ones.  We should do better, but we won’t.

Tim

dhoehna:
Hi Tim,

Sorry for the long delay in getting back to you.

I looked into the PR and have some questions and issues.
1. Log() breaks with non ASCII characters.  One example is SystemX86Fïℓè.txt.
2. The new redirection logic redirects differently for the same location.  (See example below)

Example.
This example comes from the modify file test.  In master, the following two locations redirect to the same spot. `C:\Program Files\WindowsApps\FileSystemTest_0.0.0.1_x64__8wekyb3d8bbwe\VFS\SystemX86\SystemX86Fïℓè.txt` and `C:\WINDOWS\SysWOW64\SystemX86Fïℓè.txt` both redirect to `\\?\C:\Users\dahoehna\AppData\Local\Packages\FileSystemTest_8wekyb3d8bbwe\LocalCache\Local\VFS\C$\WINDOWS\SysWOW64\SystemX86Fïℓè.txt`

On your branch, the two locations are redirected differently.  `C:\Program Files\WindowsApps\FileSystemTest_0.0.0.1_x64__8wekyb3d8bbwe\VFS\SystemX86\SystemX86Fïℓè.txt` redirects to `\\?\C:\Users\dahoehna\AppData\Local\Packages\FileSystemTest_8wekyb3d8bbwe\LocalCache\Local\Microsoft\WritablePackageRoot\PackageCache\FileSystemTest_8wekyb3d8bbwe\VFS\SystemX86\SystemX86Fïℓè.txt` and `C:\WINDOWS\SysWOW64\SystemX86Fïℓè.txt` redirects to `\\?\C:\Users\dahoehna\AppData\Local\Packages\FileSystemTest_8wekyb3d8bbwe\LocalCache\Local\Microsoft\WritablePackageRoot\PackageCache\FileSystemTest_8wekyb3d8bbwe\VFS\Windows\SysWOW64\SystemX86Fïℓè.txt`

I see what PathRedirection:ShouldRedirectImpl normalizedPath is not used anymore for the path redirection and instead ad Virtualized path is used as the basis for path redirection.  I'm confused as to why that is the case.

Because of using the virtualized path  issue number 2 crops up because both paths have a different virtualized location.

Besides those two issues the code looks good and this PR will most defiantly help PSF.

If you want, you can run the tests that come bundles with PSF to make sure changes don't break backwards compatibility.

1. In an elevated VS prompt run `BuildAll.cmd`.
2. In an elevated powershell window cd into `tests` and run `RunTests.ps1`.

To run a specific test
1. Run the above steps
2. Individual appx's will be in the `tests\scenarios\Appx` folder. 

From there you can install the individual tests and see the results from the tests and why they failed.

dhoehna:
Also.  I will handle the merge conflicts myself when the PR is ready to be merged.

TimMangan:
surfers. Summer vacation has ended and I am on the road teaching three weeks of classes. So not much time available, but I'll try to take a look.

I assume the redirection issue is ONLY that the two redirects are different from each other,  and the change to writablepackageroot is desired.  I am not sure it matters that they redirect differently as a given package has either one or the other, but I'll look to see if one path is missing a "normalization".

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

dhoehna:
Hey Tim.  No hurries, go teach your classes.  I'll be here when you get back.

TimMangan:
I looked into #1 (I can only review the source at the moment).

If you are referring to Log() in fixups/FileRedirectionFixup/PathRedirection.cpp
this is code pretty much copied from PsfLauncher/main.cpp, which looks like it has now moved to Logger.h in that module in the Microsoft main branch of the sources.  So the problem is likely bigger and wider-spread.

I don’t deal much with non-ascii character sets, but I’ll look into std::vsnprintf to see if there is a suitable replacement.

Regards,

Tim


Hi Tim,

Sorry for the long delay in getting back to you.

I looked into the PR and have some questions and issues.

  1.  Log() breaks with non ASCII characters. One example is SystemX86Fïℓè.txt.
  2.  The new redirection logic redirects differently for the same location. (See example below)

Example.
This example comes from the modify file test. In master, the following two locations redirect to the same spot. C:\Program Files\WindowsApps\FileSystemTest_0.0.0.1_x64__8wekyb3d8bbwe\VFS\SystemX86\SystemX86Fïℓè.txt and C:\WINDOWS\SysWOW64\SystemX86Fïℓè.txt both redirect to \\?\C:\Users\dahoehna\AppData\Local\Packages\FileSystemTest_8wekyb3d8bbwe\LocalCache\Local\VFS\C$\WINDOWS\SysWOW64\SystemX86Fïℓè.txt

On your branch, the two locations are redirected differently. C:\Program Files\WindowsApps\FileSystemTest_0.0.0.1_x64__8wekyb3d8bbwe\VFS\SystemX86\SystemX86Fïℓè.txt redirects to \\?\C:\Users\dahoehna\AppData\Local\Packages\FileSystemTest_8wekyb3d8bbwe\LocalCache\Local\Microsoft\WritablePackageRoot\PackageCache\FileSystemTest_8wekyb3d8bbwe\VFS\SystemX86\SystemX86Fïℓè.txt and C:\WINDOWS\SysWOW64\SystemX86Fïℓè.txt redirects to \\?\C:\Users\dahoehna\AppData\Local\Packages\FileSystemTest_8wekyb3d8bbwe\LocalCache\Local\Microsoft\WritablePackageRoot\PackageCache\FileSystemTest_8wekyb3d8bbwe\VFS\Windows\SysWOW64\SystemX86Fïℓè.txt

I see what PathRedirection:ShouldRedirectImpl normalizedPath is not used anymore for the path redirection and instead ad Virtualized path is used as the basis for path redirection. I'm confused as to why that is the case.

Because of using the virtualized path issue number 2 crops up because both paths have a different virtualized location.

Besides those two issues the code looks good and this PR will most defiantly help PSF.

If you want, you can run the tests that come bundles with PSF to make sure changes don't break backwards compatibility.

  1.  In an elevated VS prompt run BuildAll.cmd.
  2.  In an elevated powershell window cd into tests and run RunTests.ps1.

To run a specific test

  1.  Run the above steps
  2.  Individual appx's will be in the tests\scenarios\Appx folder.

From there you can install the individual tests and see the results from the tests and why they failed.

TimMangan:
Daren,

On the following:
 2.  The new redirection logic redirects differently for the same location.

I don't remember why the normalize was removed, likely I was solving a different issue.

A given package should only have a file in one or the other of those two locations.  So where the file is redirected to within the package shouldn't matter.  Thinking ahead to connection groups (or whatever it will be called in the future when we have two packages running together), if one package had the file referenced one way and another had it referenced another way and we wanted FRF redirection, perhaps that they _should_  be redirected differently?    I wouldn't want to put that logic back unless we are sure there is a good reason.

dhoehna:
Hi Tim,

Are you still touring?  If not, how was it?

You are very correct that apps won't usually write to those locations in one package.  

After talking with co-workers I realized that, as long as PSF is redirecting the paths it is told to redirect, it does not matter where PSF redirects the paths.

You don't need to keep the normalized path.  But, can you please remove the code if it is no longer being used?

One problem with changing the redirection logic is that the tests need to change to accommodate your change. I'll be talking with my co-workers to figure out the best way to introduce the test changes into the repo.

dhoehna:
Hey Tim,

Another note.  If it is easier for you, you can remove the default redirection to writable package root.  If someone needs the writable package root we can use the redirectedBasePath to force the redirection tot he writable package root.

TimMangan:
Offhand I’d rather just get what we have in for now.  I did see an issue with the extra long filepaths on redirection getting truncated at 256 in debug console displays (DebugView), but PsfMonitor shouldn’t be affected.

I have figured out non-ascii for FileRedirectionFixup and am testing that now.  I’m concerned I might also have to hit other logging code too…

Regards,

Tim

TimMangan:
I am back home for a couple of weeks (then off to Ignite).  Good to hear that normalized path removal is OK.  I’ll remove it from the code when I make logging changes for non-ascii filenames.

Regards,

Tim

TimMangan:
Updates to address non-ascii file logging done.

dhoehna:
Hi Tim,

The latest commit you have is on June 20th.  can you please push your changes so I can go over them?

Thanks,

Darren.

dhoehna:
Hey Tim,

I'll look into the PR and fix the merge conflicts.  When it is done I'll merge them into this branch and complete the pull request.

TimMangan:
Thank you!

dhoehna:
You are welcome.

FYI.  I'm going to be fixing the tests to work with the new redirection logic.  One part of my PR will be documentation on how to run the tests and to make sure all the tests pass before doing a PR.

Any PR's after this will require all tests to pass.

PR #38, opened 2019-04-17
Making path lowercase in PathRedirection

dhoehna:
.find is case sensitive.  To make sure we are comparing the correct path regardless of case, we will lower-case the packageroot path and the de-virtualized path.

### DIFF
--- a/fixups/FileRedirectionFixup/PathRedirection.cpp
+++ b/fixups/FileRedirectionFixup/PathRedirection.cpp
@@ -33,8 +33,8 @@ void InitializePaths()
     // For path comparison's sake - and the fact that std::filesystem::path doesn't handle (root-)local device paths all
     // that well - ensure that these paths are drive-absolute
     auto packageRootPath = std::wstring(::PSFQueryPackageRootPath());
-    if (auto pathType = psf::path_type(packageRootPath.c_str());
-        (pathType == psf::dos_path_type::root_local_device) || (pathType == psf::dos_path_type::local_device))
+    auto pathType = psf::path_type(packageRootPath.c_str());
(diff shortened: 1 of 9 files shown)


fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -32,14 +32,17 @@ void InitializePaths()
 {
     // For path comparison's sake - and the fact that std::filesystem::path doesn't handle (root-)local device paths all
     // that well - ensure that these paths are drive-absolute
-    auto packageRootPath = ::PSFQueryPackageRootPath();
-    if (auto pathType = psf::path_type(packageRootPath);
+    auto packageRootPath = std::wstring(::PSFQueryPackageRootPath());
+    if (auto pathType = psf::path_type(packageRootPath.c_str());
```
jolonghi: Should the if statement be on the next line and have the assignment here always?

fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -351,7 +354,12 @@ std::wstring RedirectedPath(const normalized_path& deVirtualizedPath, bool ensur
     //a path that contains the package family name and not the package full name.
     std::wstring result;
     bool shouldredirectToPackageRoot = false;
-    if (deVirtualizedPath.full_path.find(g_packageRootPath) != std::wstring::npos)
+    auto deVirtualizedFullPath = deVirtualizedPath.full_path;
+
+    //Lowercase the devirtualized full path because .find is case-sensitive.
+    transform(deVirtualizedFullPath.begin(), deVirtualizedFullPath.end(), deVirtualizedFullPath.begin(),        towlower);
```
jolonghi: nit: spacing before towlower

PR #42, opened 2019-05-09
Powershell Script support

dhoehna:
Adding the ability to run one powershell script before the exe runs and one powershell script after the exe runs.


include/psf_runtime.h
```diff
@@ -42,6 +42,9 @@ PSFAPI const psf::json_value* __stdcall PSFQueryConfigRoot() noexcept;
 PSFAPI const psf::json_object* __stdcall PSFQueryAppLaunchConfig(_In_ const wchar_t* applicationId, bool verbose) noexcept;
 PSFAPI const psf::json_object* __stdcall PSFQueryCurrentAppLaunchConfig(bool verbose) noexcept;
 PSFAPI const psf::json_object* __stdcall PSFQueryAppMonitorConfig() noexcept;
+PSFAPI const psf::json_object* __stdcall PSFQueryCurrentAppLaunchConfig(bool verbose) noexcept;
```
dunhor: Unintentional? Declared above

PsfLauncher/ErrorInformation.cpp
```diff
@@ -0,0 +1,76 @@
...
+    this->exeName = exeName;
+
+    this->errorMessage = widen(std::system_category().message(this->errorNumber));
+    this->errorMessage.resize(this->errorMessage.length() - 3);
+    this->isThereAnError = true;
+}
+
+ErrorInformation::ErrorInformation()
+{
+    this->customMessage << L"";
```
dunhor: Isn't this a no-op?

PsfLauncher/ErrorInformation.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+ErrorInformation &ErrorInformation::operator=(const ErrorInformation &toAssign)
+{
+    if (this == &toAssign)
+    {
+        return *this;
+    }
+
+    this->customMessage << toAssign.customMessage.str();
```
dunhor: This is appending. Don't you want to reset the string first?

PsfLauncher/ErrorInformation.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    return *this;
+}
+
+std::wstring ErrorInformation::Print()
+{
+    this->customMessage << L"\r\nMessage: " << this->errorMessage << "\r\n";
+    this->customMessage << L"Error number: " << this->errorNumber << "\r\n";
+
+    return this->customMessage.str();
```
dunhor: If this is called more than once, the output will contain duplicated output. Is this intentional? Or is it expected to never be called more than once?

Common.props
```diff
@@ -12,7 +12,7 @@
   <!-- Build with a common platform version -->
   <PropertyGroup Label="Globals">
     <!--Use RS1 SDK for compatibility with Desktop Bridge debut and build agent-->
-    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)'==''">10.0.14393.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)'==''">10.0.15063.0</WindowsTargetPlatformVersion>
```
Scottj1s: >15063 [](start = 89, length = 5)

what's the rs2 dependency?

PsfLauncher/ErrorInformation.cpp
```diff
@@ -0,0 +1,76 @@
...
+    this->customMessage << toAssign.customMessage.str();
+    this->errorMessage = toAssign.errorMessage;
+    this->isThereAnError = toAssign.isThereAnError;
+    this->errorNumber = toAssign.errorNumber;
+
+    return *this;
+}
+
+std::wstring ErrorInformation::Print()
+{
```
Scottj1s: looks like the effective public accessor for customMessage, which could otherwise just be declared as a std::wstring

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+    LPCWSTR CurrentDirectory;
+};
+
+//Forward declarations
+ErrorInformation StartProcess(ExecutionInformation execInfo, int cmdShow, bool runInAppContainer);
+int launcher_main(PWSTR args, int cmdShow) noexcept;
+ErrorInformation RunScript(const psf::json_object * scriptInformation, std::filesystem::path packageRoot, LPCWSTR dirStr, int cmdShow);
+ErrorInformation GetAndLaunchMonitor(const psf::json_object * monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr);
+ErrorInformation LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t * executable, const wchar_t * arguments, bool wait, bool asadmin, int cmdShow, LPCWSTR dirStr);
+static inline bool check_suffix_if(iwstring_view str, iwstring_view suffix);
```
dunhor: >iwstring_view str, iwstring_view suffix [](start = 35, length = 39)

Pass by reference

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+    std::wstring CommandLine;
+    LPCWSTR CurrentDirectory;
+};
+
+//Forward declarations
+ErrorInformation StartProcess(ExecutionInformation execInfo, int cmdShow, bool runInAppContainer);
+int launcher_main(PWSTR args, int cmdShow) noexcept;
+ErrorInformation RunScript(const psf::json_object * scriptInformation, std::filesystem::path packageRoot, LPCWSTR dirStr, int cmdShow);
+ErrorInformation GetAndLaunchMonitor(const psf::json_object * monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr);
+ErrorInformation LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t * executable, const wchar_t * arguments, bool wait, bool asadmin, int cmdShow, LPCWSTR dirStr);
```
dunhor: >std::filesystem::path packageRoot [](start = 43, length = 33)

Seems like this should be passed by reference

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+static inline bool check_suffix_if(iwstring_view str, iwstring_view suffix);
+void LogStringW(const char* name, const wchar_t* value);
+void LogString(const char* name, const char* value);
+void Log(const char* fmt, ...);
+ErrorInformation StartWithShellExecute(std::filesystem::path packageRoot, std::filesystem::path exeName, std::wstring exeArgString, LPCWSTR dirStr, int cmdShow);
```
dunhor: >std::filesystem::path packageRoot, std::filesystem::path exeName, std::wstring exeArgString [](start = 39, length = 91)

Also seems like these should be passed by reference

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+int __stdcall wWinMain(HINSTANCE, HINSTANCE, PWSTR args, int cmdShow)
 {
-    Log("\t%s=%ls\n", name, value);
+    return launcher_main(args, cmdShow);
 }
+
 int launcher_main(PWSTR args, int cmdShow) noexcept try
 {
+    //TEST
+    MessageBoxEx(NULL, L"In here", L"In here", 0, 0);
```
dunhor: ?

PsfLauncher/ErrorInformation.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+#include <string>
+#include <windows.h>
+#include <iostream>
+#include "utilities.h"
+class ErrorInformation
```
Scottj1s: nit: a little white space might improve readability

PsfLauncher/ErrorInformation.h
```diff
@@ -0,0 +1,28 @@
...
+    void AddExeName(std::wstring exeNameToAdd);
+
+private:
+    std::wostringstream customMessage;
+    DWORD errorNumber;
+    std::wstring exeName;
+    std::wstring errorMessage;
+    bool isThereAnError;
+};
+
```
Scottj1s: this whole class could be defined inline (delete the cpp)

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
 #include <psf_runtime.h>
 #include <shellapi.h>
 #include <combaseapi.h>
+#include <ppltasks.h>
+#include <ShObjIdl.h>
+#include "ErrorInformation.h"
 
-using namespace std::literals;
+//These two macros don't exist in RS1.  Define them here to prevent build
```
Scottj1s: >RS1 [](start = 34, length = 3)

but common.props is RS2 ...

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
-        va_end(args2);
-    }
+    LPCWSTR ApplicationName;
+    std::wstring CommandLine;
+    LPCWSTR CurrentDirectory;
+};
+
+//Forward declarations
+ErrorInformation StartProcess(ExecutionInformation execInfo, int cmdShow, bool runInAppContainer);
+int launcher_main(PWSTR args, int cmdShow) noexcept;
```
Scottj1s: >PWSTR [](start = 18, length = 5)

PCWSTR

PsfLauncher/main.cpp
Scottj1s: >asadmin [](start = 149, length = 7)

asAdmin (camel case)

PsfLauncher/main.cpp
Scottj1s: lots of pointers - use references where possible

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+ErrorInformation RunScript(const psf::json_object * scriptInformation, std::filesystem::path packageRoot, LPCWSTR dirStr, int cmdShow);
+ErrorInformation GetAndLaunchMonitor(const psf::json_object * monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr);
+ErrorInformation LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t * executable, const wchar_t * arguments, bool wait, bool asadmin, int cmdShow, LPCWSTR dirStr);
+static inline bool check_suffix_if(iwstring_view str, iwstring_view suffix);
+void LogStringW(const char* name, const wchar_t* value);
```
Scottj1s: >LogStringW [](start = 5, length = 10)

could just make a LogString overload

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
-        count = std::vsnprintf(str.data(), str.size() + 1, fmt, args2);
-        assert(count >= 0);
-        va_end(args2);
-    }
+    LPCWSTR ApplicationName;
+    std::wstring CommandLine;
+    LPCWSTR CurrentDirectory;
+};
+
+//Forward declarations
```
Scottj1s: use noexcept where possible

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
 
-    str.resize(count);
-    ::OutputDebugStringA(str.c_str());
-}
-void LogString(const char* name, const char* value)
-{
-    Log("\t%s=%s\n", name, value);
-}
-void LogStringW(const char* name, const wchar_t* value)
+int __stdcall wWinMain(HINSTANCE, HINSTANCE, PWSTR args, int cmdShow)
```
Scottj1s: noexcept

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
-        if (monitor_asadmin)
-            asadmin = monitor_asadmin->as_boolean().get();
-        if (monitor_wait)
-            wait = monitor_wait->as_boolean().get();
-        Log("\tCreating the monitor: %ls", monitor_executable->as_string().wide());
-        LaunchMonitorInBackground(packageRoot, monitor_executable->as_string().wide(), monitor_arguments->as_string().wide(), wait, asadmin);
-    }
+        if (!isPowershellInstalled)
+        {
+            ::PSFReportError(L"Powershell is not installed.  Please install powershell to run scripts in PSF");
```
Scottj1s: >Powershell [](start = 31, length = 10)

use "PowerShell" consistently in all strings/output

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
     // At least for now, configured launch paths are relative to the package root
     std::filesystem::path packageRoot = PSFQueryPackageRootPath();
-    auto exePath = packageRoot / exeName;
 
-    // Allow arguments to be specified in config.json now also.
-    std::wstring cmdLine = L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args;
-
-    if (monitor != nullptr )
+    ErrorInformation error;
+    bool isPowershellInstalled = CheckIfPowerShellIsInstalled();
```
Scottj1s: the error reporting and returning seems kind of scattered.  consider making ErrorInformation and RAII that condidtionally calls PSFReportError on destruction.  Then your code either stashes error string/code or checks error code.

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+
+    return 0;
 }
 catch (...)
 {
     ::PSFReportError(widen(message_from_caught_exception()).c_str());
     return win32_from_caught_exception();
+}
 
+ErrorInformation RunScript(const psf::json_object * scriptInformation, std::filesystem::path packageRoot, LPCWSTR dirStr, int cmdShow)
```
Scottj1s: >* [](start = 50, length = 1)

&

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+{
+    //Script arguments are optional.
+    auto scriptArgumentsJObject = scriptInformation->try_get("scriptArguments");
+    std::wstring scriptArguments;
+    if (scriptArgumentsJObject)
```
Scottj1s: not needed - just move the check below to where you append

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+    std::wstring powershellCommandString(L"powershell.exe -file ");
+
+    powershellCommandString.append(scriptPath);
+    powershellCommandString.append(L" ");
+    powershellCommandString.append(scriptArguments);
+
+    Log("Looking for the script here: ");
+    LogStringW("Script Path", scriptPath.c_str());
+    auto currentDirectory = (packageRoot / dirStr);
+    auto doesFileExist = std::filesystem::exists(currentDirectory / powershellScriptPath);
```
Scottj1s: nit: not needed

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
-        case WAIT_OBJECT_0:
-            break; // Success case... handle at the end of main
+            if (dirStr)
+            {
+                workingDirectory = (packageRoot / dirStr).native();
+            }
 
-        case WAIT_FAILED:
-            return ::GetLastError();
+            ExecutionInformation execInfo;
```
Scottj1s: consider just using aggregate init:  execInfo{ ... }

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+    auto doesFileExist = std::filesystem::exists(currentDirectory / powershellScriptPath);
+
+    if (doesFileExist)
+    {
+        ExecutionInformation execInfo;
+        execInfo.ApplicationName = nullptr;
+        execInfo.CommandLine = powershellCommandString;
+        execInfo.CurrentDirectory = currentDirectory.c_str();
+
+        //RunInAppEnviorment is optional.
```
Scottj1s: >RunInAppEnviorment [](start = 10, length = 18)

typo

PsfLauncher/main.cpp
```diff
@@ -13,223 +13,227 @@
...
+
+        return error;
+    }
 }
 
+ErrorInformation GetAndLaunchMonitor(const psf::json_object * monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr)
+{
+    bool asadmin = false;
+    bool wait = false;
+    auto monitor_executable = monitor->try_get("executable");
```
Scottj1s: casing is inconsistent - here it's snake_case, above it's camelCase

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
         }
         else
         {
-            //Log("error starting monitor using SellExecuteEx also. Error=0x%x\n", ::GetLastError());
+            std::wostringstream ss;
+            ss << L"error starting monitor using SellExecuteEx";
```
Scottj1s: no need for string stream

Scottj1s: >SellExecuteEx [](start = 49, length = 13)

typo

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+            0,
+            &AttributeListSize) == FALSE)
         {
-            if (::GetLastError() == ERROR_ELEVATION_REQUIRED)
-                Log("error starting monitor using CreateProcessW. You must specify 'monitor/asadmin' in config.json\n");
-            else
-                Log("error starting monitor using CreateProcessW. Error=0x%x\n", ::GetLastError());
+            auto err{ ::GetLastError() };
+            std::wostringstream ss;
+            ss << L"Could not initilize the proc thread attribute list.";
```
Scottj1s: >initilize [](start = 30, length = 9)

typo

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+            1,
+            0,
+            &AttributeListSize) == FALSE)
         {
-            if (::GetLastError() == ERROR_ELEVATION_REQUIRED)
-                Log("error starting monitor using CreateProcessW. You must specify 'monitor/asadmin' in config.json\n");
-            else
-                Log("error starting monitor using CreateProcessW. Error=0x%x\n", ::GetLastError());
+            auto err{ ::GetLastError() };
+            std::wostringstream ss;
```
Scottj1s: >wostringstream [](start = 17, length = 14)

no need for this

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        if (UpdateProcThreadAttribute(StartupInfoEx.lpAttributeList,
+            0,
+            PROC_THREAD_ATTRIBUTE_DESKTOP_APP_POLICY,
+            &attribute,
+            sizeof(attribute),
+            NULL,
+            NULL) == FALSE)
+        {
+            auto err{ ::GetLastError() };
+            std::wostringstream ss;
```
Scottj1s: >wostringstream [](start = 17, length = 14)

or this

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+            return error;
+        }
+
+        DWORD attribute = PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_DISABLE_PROCESS_TREE;
+        if (UpdateProcThreadAttribute(StartupInfoEx.lpAttributeList,
+            0,
+            PROC_THREAD_ATTRIBUTE_DESKTOP_APP_POLICY,
+            &attribute,
+            sizeof(attribute),
+            NULL,
```
Scottj1s: >NULL [](start = 12, length = 4)

NULL deprecated - use nullptr

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
-int __stdcall wWinMain(HINSTANCE, HINSTANCE, PWSTR args, int cmdShow)
+
+ErrorInformation StartWithShellExecute(std::filesystem::path packageRoot, std::filesystem::path exeName, std::wstring exeArgString, LPCWSTR dirStr, int cmdShow)
 {
-    return launcher_main(args, cmdShow);
+    // Non Exe case, use shell launching to pick up local FTA
+    auto nonexePath = packageRoot / exeName;
+
+    CoInitializeEx(NULL, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE);
+    SHELLEXECUTEINFO shex;
```
Scottj1s: why is this needed?

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+    memset(&shex, 0, sizeof(SHELLEXECUTEINFO));
+    shex.cbSize = sizeof(shex);
```
Scottj1s: not needed - just use {0} aggr init

PsfLauncher/main.cpp
```diff
@@ -268,43 +277,233 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+    Log("\t%s=%s\n", name, value);
+}
+void LogStringW(const char* name, const wchar_t* value)
+{
+    Log("\t%s=%ls\n", name, value);
+}
+
+bool CheckIfPowerShellIsInstalled()
+{
+    HKEY registryHandle;
```
Scottj1s: >registryHandle [](start = 9, length = 14)

this is leaked.  use RAII type library (like WIL) to close the handle

README.md
```diff
@@ -24,6 +24,118 @@ This branch has the latest code. Keep in mind that there might be features in th
 ## Contribute
 This project welcomes contributions and suggestions.  Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.microsoft.com.
 
+## Script support
```
Scottj1s: please proof read this for typos and grammar

tests/scenarios/PowershellScriptTest/AppxManifest.xml
```diff
@@ -0,0 +1,71 @@
...
+                          Description="No description entered" />
+    </Application>
+	<Application Id="PSTBothEndingFirst" Executable="PsfLauncher.exe" EntryPoint="Windows.FullTrustApplication">
+      <uap:VisualElements BackgroundColor="transparent"
+                          DisplayName="PST Both Ending Script First"
+                          Square150x150Logo="Assets\Logo150x150.png"
+                          Square44x44Logo="Assets\Logo44x44.png"
+                          Description="No description entered" />
+						  </Application>
+	<Application Id="PSTScriptWithArg" Executable="PsfLauncher.exe" EntryPoint="Windows.FullTrustApplication">
```
Scottj1s: mix of spaces and tabs in xml and json files - do you care?

PsfLauncher/main.cpp
dhoehna: I even put //TEST to remind myself to take it out.  Good catch.

PsfLauncher/ErrorInformation.h
dhoehna: Why should I define this inline, and, what is the reason for doing so?

Also, what do you mean by "Inline" defining all the logic in the .h file?

PsfLauncher/main.cpp
dhoehna: I'm confused by this comment.  What does common.props have to do with anything?

PsfLauncher/main.cpp
dhoehna: There are a lot of pointers.  I changed the C-style strings from * to [] but I am not going to change the json_objects to references.  At least, not for this PR.  The reason why is config.cpp returns pointers.  I could make them return a reference instead, but what would I return in case the object could not be found?

I could just make an empty json_object, but I don't know how that would affect the rest of the program.

So, right now, I won't change it.  But, that is something I can change in the future.

PsfLauncher/main.cpp
dhoehna: Why would I add noexcept?  From what I've read noexcept tells runtime that no exception will be thrown from this function and if there is one std::terminate is called instead.

PsfLauncher/main.cpp
dhoehna: What I'm trying to get here is to have one object that holds all the error information.  Thus the class ErrorInformation.  

I think the reason why it looks so scattered is because I promise that the ending script will run regardless if there is an exception, but only if the starting script ran.

Also, because I now have ErrorInformation, I needed a default ErrorInformation in case there was no error.

What is RAII?

PsfLauncher/main.cpp
dhoehna: I made a comment about this pointer in another comment.  The api in config.cpp returns a pointer to a json_object and I don't want to change the underlying implementation in this PR because I don't know what the far reaching impacts of the change are.

PsfLauncher/main.cpp
dhoehna: How would I do that?  A long ternary if statement?

I understand where you want me to put it.  Where I do the appending, but, I need to check to see if the json_oibject exists before I use it.  So, I need to first grab the json_object, then check to see if it exists.  I don't know how I would reduce it.

PsfLauncher/main.cpp
dhoehna: Right.  I don't huh.  Because if CreateProcess fails because it can't find the Script PSFReportError throws up a message box with the message.

PsfLauncher/main.cpp
dhoehna: Sure.  I'm not used to doing an aggregated init for my own objects, but it does look nicer.

Do you know if making an aggregated init is faster?

dhoehna: Wait.  No.  I'm confused now.  Are you suggesting to write the code like this

`ExecutionInformation execInfo {exePath.c_str(), cmdLine}` instead?

PsfLauncher/main.cpp
dhoehna: Fixed.

PsfLauncher/ErrorInformation.h
Scottj1s: yes, all in the .h:
- more efficient (inline methods)
- fewer source files
- easier to read/maintain

PsfLauncher/main.cpp
Scottj1s: that's where you're setting RS2 (I suggest you move it back to RS1)

PsfLauncher/main.cpp
Scottj1s: I'm referring to what a function *takes* (not what it returns).  If it never expects a null value, than reference is a stronger contract and avoids the risk, param validation, etc.

PsfLauncher/main.cpp
dhoehna: Good catch.  I went ahead and turned all but one instance of ss into a std::wstring.

PsfLauncher/main.cpp
Scottj1s: it also tells calling functions the same thing, so they in turn can omit catch handlers.  in other words, the runtime only drops in the std::terminate if you're calling functions that are not noexcept, which is suboptimal.  so it has a viral effect and you should get in the habit of applying it everywhere possible.

PsfLauncher/main.cpp
dhoehna: Fixed.

PsfLauncher/main.cpp
Scottj1s: RAII: look it up  ;-)

I have no problem with the need for ErrorInfomation, just it's error-prone usage pattern, which RAII would help.

PsfLauncher/main.cpp
dhoehna: fixed

PsfLauncher/main.cpp
dhoehna: Addressed in another comment.  Changed to std::wstring.

PsfLauncher/main.cpp
Scottj1s: again, return values are fine - params are what I'm referring to

PsfLauncher/main.cpp
dhoehna: Oh.  NULL is actually deprecated?  Can do.

PsfLauncher/main.cpp
Scottj1s: nicer and faster (as you're avoiding double-initialization, which is the same reason you should prefer constructor init-lists)

PsfLauncher/main.cpp
Scottj1s: auto scriptArgumentsJObject = scriptInformation->try_get("scriptArguments");
    if (scriptArgumentsJObject)
{
powershellCommandString.append(scriptArgumentsJObject->as_string().wide(););
}

PsfLauncher/main.cpp
dhoehna: Because the monitor program will run on a separate thread.  This is used only for PSFFixupMonitor.  But, now that I'm thinking about it, we only need it for the monitor.  I'm going to move this up to main and only invoke CoInitilizeEx if we have a monitor we need to run.

PsfLauncher/main.cpp
Scottj1s: https://osgwiki.com/wiki/WIL_Resource_Wrappers_(RAII)

(see reg handle comment below)

PsfLauncher/main.cpp
dhoehna: You mean like

`SHELLEXECUTEINFOEX shex = {0}`?

Also, why inline?

PsfLauncher/main.cpp
dhoehna: Oh, right.  Didn't close the handle.

WIL.  Are you taling about [this?](https://github.com/microsoft/wil)

tests/scenarios/PowershellScriptTest/AppxManifest.xml
dhoehna: Didn't even notice that.  Fixed

PsfLauncher/main.cpp
jolonghi: You can even just do SHELLEXECUTEINFOEX shex{};

tests/TestRunner/main.cpp
```diff
@@ -30,7 +30,12 @@ static constexpr const wchar_t* g_applications[] =
     L"CompositionTest_8wekyb3d8bbwe!Fixed",
     L"FileSystemTest_8wekyb3d8bbwe!Fixed",
     L"LongPathsTest_8wekyb3d8bbwe!Fixed",
-    L"WorkingDirectoryTest_8wekyb3d8bbwe!Fixed"
+    L"PowershellScriptTest_8wekyb3d8bbwe!PSTNoScripts",
+    L"PowershellScriptTest_8wekyb3d8bbwe!PSTOnlyStart",
+    L"PowershellScriptTest_8wekyb3d8bbwe!PSTOnlyEnd",
+    L"PowershellScriptTest_8wekyb3d8bbwe!PSTBothStartingFirst",
+    L"PowershellScriptTest_8wekyb3d8bbwe!PSTBothEndingFirst",
+    L"PowershellScriptTest_8wekyb3d8bbwe!PSTScriptWithArg"
```
jolonghi: Do we care about tests running in and out of the container?

dhoehna: Right.  I should add those.

PsfLauncher/ErrorInformation.h
dhoehna: fixed.

PsfLauncher/ErrorInformation.h
dhoehna: Done.

PsfLauncher/main.cpp
dhoehna: Done.

PsfLauncher/main.cpp
dhoehna: Done.

README.md
dhoehna: Done.

PsfLauncher/main.cpp
```diff
@@ -71,185 +64,199 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
 
     // At least for now, configured launch paths are relative to the package root
     std::filesystem::path packageRoot = PSFQueryPackageRootPath();
-    auto exePath = packageRoot / exeName;
 
-    // Allow arguments to be specified in config.json now also.
-    std::wstring cmdLine = L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args;
+    ErrorInformation error;
+    bool isPowershellInstalled = false;
+    error = CheckIfPowershellIsInstalled(isPowershellInstalled);
```
Scottj1s: nit: define and init in one line

PsfLauncher/ErrorInformation.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
+#include <string>
+#include <windows.h>
+#include <iostream>
+#include "utilities.h"
+
+class ErrorInformation
+{
+public:
+    ErrorInformation() : customMessage(L""), errorMessage(L""), isThereAnError(false)
```
dunhor: >customMessage(L""), errorMessage(L""), isThereAnError(false) [](start = 25, length = 60)

nit: use member initializers. E.g. `bool isThereAnError = false`. Also `std::basic_string` is initialized to an empty string by default; no need to explicitly do it

PsfLauncher/ErrorInformation.h
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        this->customMessage.clear();
+        this->customMessage = toAssign.customMessage;
+        this->errorMessage = toAssign.errorMessage;
+        this->isThereAnError = toAssign.isThereAnError;
+        this->errorNumber = toAssign.errorNumber;
+
+        return *this;
+    }
```
dunhor: Neither of these copy over `exeName` - is this intentional? If it's not intentional, then neither of these functions are necessary since the compiler will generate the equivalent for you

dunhor: Though `exeName` doesn't seem to be used for anything? Could also remove it, in which case neither of these functions are necessary

---
In reply to: [285251717](https://github.com/microsoft/MSIX-PackageSupportFramework/pull/42#discussion_r285251717) [](ancestors = 285251717)

tests/scenarios/PowershellScriptTest/config.json
```diff
@@ -0,0 +1,94 @@
...
+        },
+        {
+            "executable": ".*",
+            "fixups": [
+                {
+                    "dll": "MultiByteToWideCharTestFixup.dll"
+                },
+                {
+                    "dll": "CompositionTestFixup.dll"
+                }
```
dunhor: Are these necessary?

tests/TestRunner/main.cpp
dunhor: Nothing actually validates the script outputs, right? So is the idea here just that if the exit code is zero we assume that the scripts ran successfully? Perhaps considering doing something in the scripts that has a side effect that's observable that the test executable can check for and fail if it wasn't done. That'll only really test the 'start' script, but perhaps that's good enough

PsfLauncher/ErrorInformation.h
dhoehna: Thank you for pointing that out.  I added exeName.

tests/TestRunner/main.cpp
dhoehna: You are correct.  There is no way to tell if there was an error inside a script. 

I decided not to check for errors inside scripts because the script authors might expect an error to happen.  This would happen in case the script writes to a file that exists and if the file exists to not do anything.  

Because people can expect errors to happen in a script I don't test for it.

tests/scenarios/PowershellScriptTest/config.json
dhoehna: no.

PsfLauncher/ErrorInformation.h
```diff
@@ -0,0 +1,93 @@
...
+        return this->errorNumber;
+    }
+
+    void AddExeName(std::wstring exeNameToAdd)
+    {
+        this->exeName = exeNameToAdd;
+    }
+
+private:
+    std::wstring customMessage;
```
Scottj1s: >customMessage [](start = 17, length = 13)

still don't understand why this is a member:
values captured by constructors are always overwritten
value returned by Print is never cached

PsfLauncher/main.cpp
Scottj1s: define & init in one line

PsfLauncher/main.cpp
```diff
@@ -71,185 +64,199 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
-    auto monitor = PSFQueryAppMonitorConfig();
-
 
     // At least for now, configured launch paths are relative to the package root
     std::filesystem::path packageRoot = PSFQueryPackageRootPath();
-    auto exePath = packageRoot / exeName;
 
-    // Allow arguments to be specified in config.json now also.
-    std::wstring cmdLine = L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args;
+    ErrorInformation error;
```
Scottj1s: >ErrorInformation [](start = 4, length = 16)

For future code, consider an exception model where you can simply construct-and-throw the necessary bits of info.  Then you don't need to see the clutter of if checks around IsThereAnError, PSFReportError calls, and returns.  It all gets tucked away so you can focus on your business logic.

PsfLauncher/main.cpp
```diff
@@ -71,185 +64,199 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
+    std::wstring powershellCommandString(L"PowerShell.exe -file ");
+
+    std::wstring scriptPath = scriptInformation.get("scriptPath").as_string().wide();
+    powershellCommandString.append(scriptPath);
+    powershellCommandString.append(L" ");
+
+    //Script arguments are optional.
+    auto scriptArgumentsJObject = scriptInformation.try_get("scriptArguments");
+    std::wstring scriptArguments;
+    if (scriptArgumentsJObject)
```
Scottj1s: unused

PsfLauncher/main.cpp
```diff
@@ -71,185 +64,199 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
+    {
+        powershellCommandString.append(scriptArgumentsJObject->as_string().wide());
+    }
+
+    auto currentDirectory = (packageRoot / dirStr);
+    std::filesystem::path PowershellScriptPath(scriptPath);
```
Scottj1s: >P [](start = 26, length = 1)

use consistent casing

PsfLauncher/main.cpp
```diff
@@ -71,185 +64,199 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
+    auto doesFileExist = std::filesystem::exists(currentDirectory / PowershellScriptPath);
 
+    if (doesFileExist)
+    {
+        //runInVirtualEnvironment is optional.
+        auto runInVirtualEnviormentJObject = scriptInformation.try_get("runInVirtualEnvironment");
+        auto runInVirtualEnviorment = false;
```
Scottj1s: >runInVirtualEnviorment [](start = 13, length = 22)

typo

PsfLauncher/main.cpp
```diff
@@ -71,185 +64,199 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
+        errorMessage.append(L" can't be found");
+        ErrorInformation error = { errorMessage, ERROR_FILE_NOT_FOUND };
+
+        return error;
+    }
 }
 
+ErrorInformation GetAndLaunchMonitor(const psf::json_object &monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr) noexcept
+{
+    bool asadmin = false;
```
Scottj1s: use pascal case

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
         }
         else
         {
-            //Log("error starting monitor using SellExecuteEx also. Error=0x%x\n", ::GetLastError());
+            std::wstring errorString = L"error starting monitor using ShellExecuteEx";
+            auto err = ::GetLastError();
+            ErrorInformation error = { errorString, err, executable };
+
```
Scottj1s: why all the locals?  this could be nearly one line

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        execInfo.CurrentDirectory = currentDirectory.c_str();
+        ErrorInformation error = StartProcess(execInfo, cmdShow, false);
+        error.AddExeName(executable);
+
+        return error;
+    }
+
+    ErrorInformation noError;
+    return noError;
+}
```
Scottj1s: this can simply be "return {}"

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+            AttributeListSize
+        );
+
+        if (InitializeProcThreadAttributeList(startupInfoEx.lpAttributeList,
+            1,
+            0,
+            &AttributeListSize) == FALSE)
+        {
+            auto err{ ::GetLastError() };
+            std::wstring errorString = L"Could not initilize the proc thread attribute list.";
```
Scottj1s: typo

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        if (waitResult == WAIT_OBJECT_0)
+        {
+            ErrorInformation noError;
+            return noError;
         }
         else
         {
-            if (wait)
-                WaitForSingleObject(processInfo.hProcess, INFINITE);
+            auto err{ ::GetLastError() };
```
Scottj1s: >::GetLastError() [](start = 22, length = 16)

the comment indicates propagating the exit code, in which case you'd want this:
[https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess](https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess)

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+
+        if (execInfo.ApplicationName)
+        {
+            ss << L"ERROR: Failed to create a process for " << execInfo.ApplicationName << " ";
+        }
+        else
+        {
+            std::wstring applicationNameForError;
+            applicationNameForError = execInfo.CommandLine.substr(0, execInfo.CommandLine.find(' '));
+            ss << L"ERROR: Failed to create a process for " << applicationNameForError << " ";
```
Scottj1s: what if it's a LFN with spaces?

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        return error;
+    }
 }
-int __stdcall wWinMain(HINSTANCE, HINSTANCE, PWSTR args, int cmdShow)
+
+ErrorInformation StartWithShellExecute(std::filesystem::path packageRoot, std::filesystem::path exeName, std::wstring exeArgString, LPCWSTR dirStr, int cmdShow) noexcept
 {
-    return launcher_main(args, cmdShow);
+    // Non Exe case, use shell launching to pick up local FTA
+    auto nonexePath = packageRoot / exeName;
```
Scottj1s: >nonexePath [](start = 9, length = 10)

pascal casing

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        auto err{ ::GetLastError() };
+        std::wstring errorString = L"ERROR: Failed to create detoured shell process";
+
+        ErrorInformation error = { errorString, err };
+        return error;
+    }
+
+    ErrorInformation noError;
+    return noError;
 }
```
Scottj1s: return {}

PsfLauncher/UniqueHkey.h
```diff
@@ -0,0 +1,40 @@
...
+    PHKEY operator&()
+    {
+        return (PHKEY)&key;
+    }
+
+    void Set(HKEY value)
+    {
+        key = value;
+    }
+
```
Scottj1s: this too runs the risk of leaking a handle.  std::unique_ptr is a good pattern to follow

PsfLauncher/UniqueHkey.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+#include <windef.h>
+#include <handleapi.h>
+#include <winreg.h>
+
+class UniqueHKEY
```
Scottj1s: >UniqueHKEY [](start = 6, length = 10)

fyi - wil is open source now:
[https://github.com/microsoft/wil/](https://github.com/microsoft/wil/)

README.md
```diff
@@ -24,6 +24,99 @@ This branch has the latest code. Keep in mind that there might be features in th
...
+PSF allows one powershell script to be ran before an exe runs, and one powershell script to be ran after the exe runs.
```
Scottj1s: >powershell [](start = 15, length = 10)

PowerShell

Scottj1s: >ran [](start = 39, length = 3)

run

Scottj1s: >ran [](start = 95, length = 3)

run

PsfLauncher/main.cpp
```diff
@@ -268,43 +275,247 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+
+ErrorInformation CheckIfPowershellIsInstalled(bool& isPowershellInstalled) noexcept
+{
+    UniqueHKEY registryKey;
+    DWORD statusOfRegistryKey;
+    LSTATUS createResult = RegCreateKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\PowerShell\\1", 0, nullptr, REG_OPTION_NON_VOLATILE, KEY_READ, nullptr, &registryKey, &statusOfRegistryKey);
+
+    if (createResult != ERROR_SUCCESS)
+    {
+        ErrorInformation error = { L"Error with getting the key to see if PowerShell is installed. ", (DWORD)createResult };
```
Scottj1s: >ErrorInformation error =  [](start = 8, length = 25)

just return { ... }

README.md
```diff
@@ -24,6 +24,99 @@ This branch has the latest code. Keep in mind that there might be features in th
...
+| Key name                | Value type | Required?             |
+|-------------------------|------------|-----------------------|
+| scriptPath              | string     | Yes                   |
+| scriptArguments         | string     | No                    |
+| runInVirtualEnvironment | boolean    | No (defaults to true) |
+ 
+ #### Key descriptions
+ 1. scriptPath: The path to the script including the name and extension.  The Path starts from the root directory of the application.
+ 2. scriptArguments: Space delimited argument list.  The format is the same for a powershell script call.  This string gets appended to scriptPath to make a valid powershell.exe call.
+ 3. runInVirtualEnvironment: If the script should run in the same virtual enviorment that the packaged exe runs in.
```
Scottj1s: >enviorment [](start = 74, length = 10)

typo

PsfLauncher/ErrorInformation.h
dhoehna: You are correct.  customMessage is to hold a custom error message when ErrorInformation was made.  I changed print to build a new std::wstring inside of print.  That way the custom message of the error is preserved.

PsfLauncher/main.cpp
dhoehna: I don't understand, is something wrong with defining and initializing a wstring on the same line?

PsfLauncher/main.cpp
dhoehna: Yeah, as I was in the middle of adding ErrorInformation, I realized that everything had to follow the "Check then throw" pattern.

Instead of doing a "Check and throw" do more something like, construct an ErrorInformation where the error is taking place and throwing there?

PsfLauncher/main.cpp
dhoehna: Removed.

PsfLauncher/main.cpp
dhoehna: Fixed.

PsfLauncher/main.cpp
dhoehna: Fixed

PsfLauncher/main.cpp
dhoehna: Fixed.

PsfLauncher/main.cpp
dhoehna: Fixed.

tests/TestRunner/main.cpp
dunhor: Okay, I'll rephrase it this way: _what are you actually testing here_? I figured that the idea was to test that the scripts were actually run, but they don't appear to be doing that. AFAICT they just run and succeed pretty unconditionally. So, what's the point of even having these tests then?

PsfLauncher/main.cpp
dhoehna: I see.  Potentially, the wait could fail, but the process could have exited successfully and completed it's work, correct?

PsfLauncher/main.cpp
dhoehna: This would fail.

PsfLauncher/main.cpp
dhoehna: Fixed

PsfLauncher/main.cpp
dhoehna: Fixed

PsfLauncher/UniqueHkey.h
dhoehna: Added wil::uniquehkey

README.md
dhoehna: Fixed.

dhoehna: Fixed.

PsfLauncher/main.cpp
dhoehna: Is this for styling, or for performance?

README.md
dhoehna: fixed.

PsfLauncher/main.cpp
dhoehna: I did fix it by the way.

PsfLauncher/main.cpp
dhoehna: I changed it one line with return {...}

But, the answer to why is because I didn't know you could do it in one line.

tests/TestRunner/main.cpp
dhoehna: To test the config file.  I want to make sure each script will run correctly based on different configuration files.

Example.  Having an ending script but no starting script.

dunhor: > I want to make sure each script will run correctly

Right. I'm saying that you are _not_ doing that. E.g. if the script configuration were ignored (or malformed, etc.) these would still pass. All that it validates is that the executable is run; there's no validation that any scripts were actually run AFAICT

dhoehna: I see.  You are correct.  I'm not testing if any scripts were ran.

I'll make the scripts write a sentinel file and have the tests verify that the file exists.

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
         }
         else
         {
-            //Log("error starting monitor using SellExecuteEx also. Error=0x%x\n", ::GetLastError());
+            auto err = ::GetLastError();
+            ErrorInformation error = { L"error starting monitor using ShellExecuteEx", err, executable };
+        }
+
+        DWORD exitCode = ERROR_SUCCESS;
```
Scottj1s: this should be inside the if(ShellExecuteEx(...)) above

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
         {
-            //Log("error starting monitor using SellExecuteEx also. Error=0x%x\n", ::GetLastError());
+            auto err = ::GetLastError();
+            ErrorInformation error = { L"error starting monitor using ShellExecuteEx", err, executable };
+        }
+
+        DWORD exitCode = ERROR_SUCCESS;
+        if (!GetExitCodeProcess(shExInfo.hProcess, &exitCode))
+        {
+            return { L"Could not get the exit process for " + std::wstring(executable), GetLastError() };
```
Scottj1s: Could not get the exit code ...

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        }
+        else
+        {
+            if (exitCode != ERROR_SUCCESS)
+            {
+                return { L"The process " + std::wstring(executable) + L" exited with with an unsuccessful code.", exitCode };
```
Scottj1s: a non-zero process exit code does not necessarily imply failure (it's up to the process) - I'd just omit "unsuccessful"

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        (LPSTARTUPINFO)&startupInfoEx,
+        &processInfo))
+    {
+        // Propagate exit code to caller, in case they care
+        DWORD waitResult = ::WaitForSingleObject(processInfo.hProcess, INFINITE);
+
+        if (waitResult != WAIT_OBJECT_0)
+        {
+            auto err{ ::GetLastError() };
+            return { L"Waiting for the application terminated unexpectedly.", err };
```
Scottj1s: maybe "Wait operation failed unexpectedly." ?

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+    if (GetExitCodeProcess(processInfo.hProcess, &exitCode))
+    {
+        if (exitCode != ERROR_SUCCESS)
+        {
+            return { L"ERROR: Application failed for " + GetApplicationNameFromExecInfo(execInfo), exitCode };
+        }
+    }
+    else
+    {
+        return { L"Could not get the exit code process for " + GetApplicationNameFromExecInfo(execInfo), ::GetLastError() };
```
Scottj1s: just "exit code"

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+
+        return { ss.str(), err };
+    }
+
+    DWORD exitCode = ERROR_SUCCESS;
+    if (GetExitCodeProcess(processInfo.hProcess, &exitCode))
+    {
+        if (exitCode != ERROR_SUCCESS)
+        {
+            return { L"ERROR: Application failed for " + GetApplicationNameFromExecInfo(execInfo), exitCode };
```
Scottj1s: again, can't assume non-zero exit code is a failure

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+    if (GetExitCodeProcess(shex.hProcess, &exitCode))
+    {
+        if (exitCode != ERROR_SUCCESS)
+        {
+            return { L"Application exited and was not successful. " + exeName.native(), exitCode };
+        }
+    }
+    else
+    {
+        return { L"Could not get the exit process for " + exeName.native(), GetLastError() };
```
Scottj1s: exit code

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        auto err{ ::GetLastError() };
+        return { L"ERROR: Failed to create detoured shell process", err };
+    }
+    
+    DWORD exitCode = ERROR_SUCCESS;
+    if (GetExitCodeProcess(shex.hProcess, &exitCode))
+    {
+        if (exitCode != ERROR_SUCCESS)
+        {
+            return { L"Application exited and was not successful. " + exeName.native(), exitCode };
```
Scottj1s: consider factoring out a function for retrieving/reporting exit code, in part to standardize on message text

PsfLauncher/ErrorInformation.h
```diff
@@ -0,0 +1,99 @@
...
+        std::wstring errorToPrint;
+
+        if (!this->customMessage.empty())
+        {
+            errorToPrint.append(this->customMessage);
+            errorToPrint.append(L"\r\n");
+        }
+
+        errorToPrint.append(L"Message: " + this->errorMessage + L"\r\n");
+        errorToPrint.append(L"Error number: " + this->errorNumber);
```
dunhor: I don't think this does what you think it does. `L"text" + integer` Will increment the string pointer by the specified amount. E.g. `L"foobar" + 3` will give you `bar`, not `foobar3`. See [`std::to_string`](https://en.cppreference.com/w/cpp/string/basic_string/to_string)

PsfLauncher/ErrorInformation.h
```diff
@@ -0,0 +1,99 @@
...
+    ErrorInformation(std::wstring customMessage, DWORD errorNumber, std::wstring exeName)
+        : customMessage(customMessage)
+        , errorNumber(errorNumber)
+        , errorMessage(widen(std::system_category().message(errorNumber)))
+        , isThereAnError(true)
+        , exeName(exeName)
+    {
+    }
+
+    ErrorInformation(const ErrorInformation &toCopy)
```
dunhor: Both the copy constructor and copy assignment operator are unnecessary as the compiler will generate them for you. You can remove them

PsfLauncher/main.cpp
```diff
@@ -71,185 +66,198 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
-            return err;
-        }
+            std::wstring workingDirectory;
 
-        // Propagate exit code to caller, in case they care
-        switch (::WaitForSingleObject(processInfo.hProcess, INFINITE))
-        {
-        case WAIT_OBJECT_0:
-            break; // Success case... handle at the end of main
+            if (dirStr)
```
dunhor: You made this guaranteed to always be true

PsfLauncher/main.cpp
```diff
@@ -71,185 +66,198 @@ int launcher_main(PWSTR args, int cmdShow) noexcept try
...
+
+    if (doesFileExist)
+    {
+        //runInVirtualEnvironment is optional.
+        auto runInVirtualEnvironmentJObject = scriptInformation.try_get("runInVirtualEnvironment");
+        auto runInVirtualEnvironment = false;
+
+        if (runInVirtualEnvironmentJObject)
+        {
+            runInVirtualEnvironment = runInVirtualEnvironmentJObject->as_string().wide();
```
dunhor: `runInVirtualEnvironment` is a `bool`. Why are you assigning a string to it?

PsfLauncher/main.cpp
```diff
@@ -268,43 +276,278 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+    {
+        return { L"Could not get the exit process for " + exeName.native(), GetLastError() };
+    }
+
+    return {};
 }
+
+static inline bool check_suffix_if(iwstring_view str, iwstring_view suffix) noexcept
+{
+    if ((str.length() >= suffix.length()) && (str.substr(str.length() - suffix.length()) == suffix))
```
dunhor: `if (x) return true; else return false;` can always be collapsed to `return x`

PsfLauncher/main.cpp
dhoehna: Good idea.

PsfLauncher/ErrorInformation.h
dhoehna: I had no idea such a function existed.  Thank you.

PsfLauncher/ErrorInformation.h
dhoehna: When do I need to make a copy and assignment operator then?  I thought I had to make it for all custom objects I made.

dunhor: You need to define a custom copy constructor/assignment operator when your members' copy constructors are not good enough. E.g. if you need to do any resource management.

PsfLauncher/main.cpp
```diff
@@ -265,46 +277,259 @@ void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t
...
+        &processInfo))
+    {
+        // Propagate exit code to caller, in case they care
+        DWORD waitResult = ::WaitForSingleObject(processInfo.hProcess, INFINITE);
+
+        if (waitResult != WAIT_OBJECT_0)
+        {
+            auto err{ ::GetLastError() };
+            return { L"Waiting operation failed unexpectedly.", err };
+        }
```
Scottj1s: not propagating exit code?

dhoehna: That was an old comment from Tim.  That is removed.

PR #72, opened 2019-06-07
Ensure that directory structure for file redirection is created upfront.

jaholme:
This change ensures that the directory structure for a package utlizing write redirection is created upfront.  In a recent release packages can optionally disable AppData write virtualization, and this results in the  Packages\<PackageFamilyName>\LocalCache\Local folder structure not being created by the operating system. 

Additionally, the VFS redirection path should be relative to the package AppData folder rather than the LocalAppData known folder root.  

Also, refactors the powershell existence check to only happen if the config actually specifies to run powershell scripts. The error handling also does not consider the non-existence of the key as an error, instead it returns early that powershell is not installed.


fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -28,6 +28,25 @@ struct vfs_folder_mapping
 };
 std::vector<vfs_folder_mapping> g_vfsFolderMappings;
 
+std::filesystem::path CreateDirectoryTree(std::filesystem::path path, std::wstring_view relativePath)
```
dhoehna: This logic already exists.  GenerateRedirectedPath in PathRedirection.cpp.  If you pass in true for "ensureDIretoryExists" the method will create the directory tree.

jaholme: Replaced with std::filesystem::create_directories. GenereateRedirectedPath has a subtle difference that it doesn't create the last path fragment.

PR #77, opened 2019-07-18
Partial trust and exception fixes

jolonghi:
Cleaned up execution path of PSFLauncher/Runtime and fixed ElectronFixup for partial trust execution to work with a base HelloWorld application of the Electron framework.
 - Removed pattern of throwing access violation in configuration lookup and catching the error.
 - Moved Electronfixup to statically link library calls
 - Tested by running automated testsuite
 - Tested electronfixup on electron helloWorld 5.0.6


PsfLauncher/main.cpp
```diff
@@ -40,31 +39,23 @@ TRACELOGGING_DEFINE_PROVIDER(
...
-ErrorInformation GetAndLaunchMonitor(const psf::json_object &monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr) noexcept;
-ErrorInformation LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t executable[], const wchar_t arguments[], bool wait, bool asAdmin, int cmdShow, LPCWSTR dirStr) noexcept;
+void GetAndLaunchMonitor(const psf::json_object &monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr);
+void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t executable[], const wchar_t arguments[], bool wait, bool asAdmin, int cmdShow, LPCWSTR dirStr);
+
 static inline bool check_suffix_if(iwstring_view str, iwstring_view suffix) noexcept;
 void LogString(const char name[], const wchar_t value[]) noexcept;
 void LogString(const char name[], const char value[]) noexcept;
 void Log(const char fmt[], ...) noexcept;
-ErrorInformation StartWithShellExecute(std::filesystem::path packageRoot, std::filesystem::path exeName, std::wstring exeArgString, LPCWSTR dirStr, int cmdShow) noexcept;
```
jsidewhite: >ErrorInformation [](start = 0, length = 16)

Glad to see all these ErrorInformations gone.

PsfLauncher/main.cpp
```diff
@@ -512,7 +375,7 @@ void LogString(const char name[], const wchar_t value[]) noexcept
     Log("\t%s=%ls\n", name, value);
 }
 
-ErrorInformation CheckIfPowershellIsInstalled(bool& isPowershellInstalled) noexcept
+void CheckIfPowershellIsInstalled(bool& isPowershellInstalled)
```
jsidewhite: >bool& isPowershellInstalled [](start = 34, length = 27)

Now that this function uses exceptions, you can just return this bool instead of using a gross out-param.

PsfLauncher/main.cpp
```diff
@@ -462,15 +324,16 @@ ErrorInformation StartWithShellExecute(std::filesystem::path packageRoot, std::f
...
-        auto err{ ::GetLastError() };
-        return { L"ERROR: Failed to create detoured shell process", err };
-    }
-
-    return GetExitCodeForProcess({ exeName.native().c_str() }, shex.hProcess);
+    THROW_LAST_ERROR_IF_MSG (
+        !ShellExecuteEx(&shex),
+        "ERROR: Failed to create detoured shell process");
+
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE), shex.hProcess == INVALID_HANDLE_VALUE);
```
jsidewhite: >THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE), shex.hProcess == INVALID_HANDLE_VALUE); [](start = 4, length = 93)

Instead of throwing this, does it make more sense to throw the static_cast<int>(shex.hInstApp)?

See [https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecuteexa](https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecuteexa)

PsfLauncher/main.cpp
jolonghi: I like it, changed to return bool

---
In reply to: [313159157](https://github.com/microsoft/MSIX-PackageSupportFramework/pull/77#discussion_r313159157) [](ancestors = 313159157)

PsfLauncher/main.cpp
jolonghi: Seems like I can just get last error in this case according to the documentation. I think that is the safest thing to do here.

---
In reply to: [313160279](https://github.com/microsoft/MSIX-PackageSupportFramework/pull/77#discussion_r313160279) [](ancestors = 313160279)

PR #105, opened 2019-12-06
PsfDllLibraryFixup

TimMangan:
This PR does the following:

- Adds a new type of fixup for the PSF called PsfDllLibraryFixup.  This fixup intercepts calls to load a dll module to ensure that if the dll is located within the package it will be found.  This solves issues with repackaging traditional application installers that used a modification to the Path variable or an AppPaths registration in the Windows Registry so that the app would find it's dlls.  But it serves as a quick fix for any situation where testing shows a dll not being found when running in the container.
- Completes changes to allow the PSF to be used when the target exe is not in the root folder.
- Changes the PsfMonitor build to be bit specific (x86/x64 rather than AnyCPU). While the bitness of the monitor isn't tied to the bitness of the application, AnyCPU required the bitness of the dependency dlls to match the end-user environment, which could change based on Group Policy; thus it is better to just go bitness specific.
- PsfMonitor now includes the capture of the standard Windows Application and System logs. This helps when the application is trying to tell you what is wrong in those traditional sources.Small fixes to the FRF detected while testing specific one-off scenarios.
- Changes to FRF logging to solve some messages that didn't appear due to wide conversion issues..


dhoehna:
Can you change the PR to develop?  I've made the same mistake.  I need to see how to stop master from being the default PR target.

TimMangan:
Replacing this with a PR on the Develop branch.

PR #107, opened 2019-12-10
Replacement PR for 105 on main branch

TimMangan:
This is a replacement PR for #105 made to the main branch.

This PR does the following:
- Adds a new type of fixup for the PSF called PsfDllLibraryFixup. This fixup intercepts calls to load a dll module to ensure that if the dll is located within the package it will be found. This solves issues with repackaging traditional application installers that used a modification to the Path variable or an AppPaths registration in the Windows Registry so that the app would find it's dlls. But it serves as a quick fix for any situation where testing shows a dll not being found when running in the container.
- Completes changes to allow the PSF to be used when the target exe is not in the root folder.
- Changes the PsfMonitor build to be bit specific (x86/x64 rather than AnyCPU). While the bitness of the monitor isn't tied to the bitness of the application, AnyCPU required the bitness of the dependency dlls to match the end-user environment, which could change based on Group Policy; thus it is better to just go bitness specific.
- PsfMonitor now includes the capture of the standard Windows Application and System logs. This helps when the application is trying to tell you what is wrong in those traditional sources. 
- Small fixes to the FRF detected while testing specific one-off scenarios.
- Changes to FRF logging to solve some messages that didn't appear due to wide conversion issues.


dhoehna:
Hey Tim,

Next time you can change the PR branch without making a new PR.  If you hit "Edit" to the right of the PR name you can change the target branch.  Just an FYI.

fixups/FileRedirectionFixup/FileAttributesFixup.cpp
```diff
@@ -14,7 +14,7 @@ DWORD __stdcall GetFileAttributesFixup(_In_ const CharT* fileName) noexcept
     {
         if (guard)
         {
-            Log(L"GetFileAttributesFixup for %ls", fileName);
+            LogString(L"GetFileAttributesFixup for fileName", fileName);
```
dhoehna: Thanks for changing this to string literals.  That will help so much.

fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -244,19 +250,35 @@ void Log(const wchar_t* fmt, ...)
             va_start(args2, fmt);
             count = std::vswprintf(wstr.data(), wstr.size() + 1, fmt, args2);
             va_end(args2);
-            assert(count >= 0);
         }
         wstr.resize(count);
 #if _DEBUG
         ::OutputDebugStringW(wstr.c_str());
-		::OutputDebugStringW(L"\n");
+		//::OutputDebugStringW(L"\n");
```
dhoehna: Did you mean to leave this commented out?  If we don't need it can it be removed.

fixups/DynamicLibraryFixup/DynamicLibraryFixup.cpp
```diff
@@ -0,0 +1,142 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#include <psf_framework.h>
+
+
+#include "FunctionImplementations.h"
+#include "dll_location_spec.h"
+
+extern bool                  g_dynf_forcepackagedlluse;
```
dhoehna: Can you remove the extra space please.

dhoehna:
The code looks nice, I added a few comments.  

Two things that need to be done before the PR is accepted.
1. All the tests need to pass.  Please use RunTests.ps1 as admin to make sure all tests pass.
2. The new fixup needs some tests to make sure any future changes does not break DynamicLibraryFixup.  I can provide steps on how to add a new test.

We have a test solution under the tests directory called tests.sln.  Here is what you need to do for add tests for DynamicLibraryFixup
1. Add a new project for your fixup.  Something like `DynamicLibraryTest` should work
2. Add a main.cpp file (You can look at `PowershellScriptTest` as a template)
3. You will need to have `test_initilize`, `test_begin`, `test_end`, and `test_cleanup` calls.
4. In main.cpp you can add you tests to test the functionality of the fixup.

In addition to adding your project you need to manually add the project name you made above to some other locations.
1. g_applications in TestRunner:main.cpp.  It is in the format [ProjectName]![AppxName]
2. $packagesToUninstall.ps1 line 42.  You need to add your project name to the list to make sure the tests will uninstall the package.

That should be it.  These tests need to be green as well.

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: removed.

fixups/DynamicLibraryFixup/DynamicLibraryFixup.cpp
TimMangan: removed

TimMangan:
Working on adding the new test.
Also will remove AnyCPU builds from the buildall script as ANYCPU was removed from CetennialFixup solution.

Will report back when done.

TimMangan:
New test has been added.  Found a few things to cleanup in the FRF, including finally getting to the bottom of the problem with Log().

All Tests pass and this should be good to go?

dhoehna:
Thank you very much for your work.  It looks good.  I give it my seal of approval.

PR #135, opened 2020-03-16
Making it so Powershell starts in the container.

dhoehna:
This fixes a bug where the main application would create all processes inside a container.  This behavior is fixed.  The only process created in the container is

1. All powershell scripts.
2. The application that PSF launches.


jsidewhite:
I'm confused about why there's 17 commits - some from December.  Let's redo this PR and avoid the 'develop' branch.  Rather just merge a commit from dhoehna:fix to microsoft:master - preferably 1 squashed commit (or at least some small number of commits).

PR #136, opened 2020-03-16
Fixing bug where application created applications in the container.

dhoehna:
The last patch included a bug that would make all child processes be created in the container.

The behavior has been reverted and now only powershell.exe will be created in the container.  This fix will still include the ability for release builds to work properly as well.

### DIFF
--- a/PsfLauncher/PsfPowershellScriptRunner.h
+++ b/PsfLauncher/PsfPowershellScriptRunner.h
@@ -36,36 +36,93 @@ public:
 
         if (startScriptInformationObject)
         {
-            this->startingScriptInformation = MakeScriptInformation(startScriptInformationObject, stopOnScriptError, currentDirectory);
-            this->startingScriptInformation.doesScriptExistInConfig = true;
+            m_StartingScriptInformation = MakeScriptInformation(startScriptInformationObject, stopOnScriptError, currentDirectory);
+            m_StartingScriptInformation.doesScriptExistInConfig = true;
+            m_AttributeList.Initilize();
         }
 
         if (endScriptInformationObject)
         {
             //Ending script ignores stopOnScriptError.  Keep it the default value
-            this->endingScriptInformation = MakeScriptInformation(endScriptInformationObject, false, currentDirectory);
-            this->endingScriptInformation.doesScriptExistInConfig = true;
+            m_EndingScriptInformation = MakeScriptInformation(endScriptInformationObject, false, currentDirectory);
+            m_EndingScriptInformation.doesScriptExistInConfig = true;
 
             //Ending script ignores this value.  Keep true to make sure
             //script runs on the current thread.
-            this->endingScriptInformation.waitForScriptToFinish = true;
-            this->endingScriptInformation.stopOnScriptError = false;
+            m_EndingScriptInformation.waitForScriptToFinish = true;
+            m_EndingScriptInformation.stopOnScriptError = false;
         }
     }
 
     //RunStartingScript should return an error only if stopOnScriptError is true
     void RunStartingScript()
     {
-        RunScript(this->startingScriptInformation);
+        RunScript(m_StartingScriptInformation, m_AttributeList);
     }
 
     void RunEndingScript()
     {
-        RunScript(this->endingScriptInformation);
+        RunScript(m_EndingScriptInformation, m_AttributeList);
     }
 
 private:
-    DWORD createInContainerAttribute = 0x02;
+
+    struct MyProcThreadAttributeList
+    {
+    private:
+        // To make sure the attribute value persists we cache it here.
+        // 0x02 is equivalent to PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_DISABLE_PROCESS_TREE
+        // The documentation can be found here:
+        //https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute
+        // This attribute tells PSF to
+        // 1. Create Powershell in the same container as PSF.
+        // 2. Any windows Powershell makes will also be in the same container as PSF.
+        // This means that powershell, and any windows it makes, will have the same restrictions as PSF.
+        DWORD createInContainerAttribute = 0x02;
+        std::unique_ptr<LPPROC_THREAD_ATTRIBUTE_LIST> attributeList;
+
+    public:
+
+        LPPROC_THREAD_ATTRIBUTE_LIST get()
+        {
+            return *(attributeList.get());
+        }
+
+        void Initilize()
+        {
+            SIZE_T AttributeListSize{};
+            InitializeProcThreadAttributeList(nullptr, 1, 0, &AttributeListSize);
+
+            attributeList = 
+                std::make_unique<LPPROC_THREAD_ATTRIBUTE_LIST>((LPPROC_THREAD_ATTRIBUTE_LIST) operator new(AttributeListSize));
+
+            THROW_LAST_ERROR_IF_MSG(
+                !InitializeProcThreadAttributeList(
+                    *(attributeList.get()),
+                    1,
+                    0,
+                    &AttributeListSize),
+                "Could not initialize the proc thread attribute list.");
+
+            THROW_LAST_ERROR_IF_MSG(
+                !UpdateProcThreadAttribute(
+                    *(attributeList.get()),
+                    0,
+                    ProcThreadAttributeValue(18, FALSE, TRUE, FALSE),
+                    &createInContainerAttribute,
+                    sizeof(createInContainerAttribute),
+                    nullptr,
+                    nullptr),
+                "Could not update Proc thread attribute.");
+        }
+
+        ~MyProcThreadAttributeList()
+        {
+            DeleteProcThreadAttributeList(*(attributeList.get()));
+        }
+
+
+    };
 
     struct ScriptInformation
     {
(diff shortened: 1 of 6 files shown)


PsfLauncher/StartProcessHelper.h
```diff
@@ -58,9 +27,11 @@ HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR curren
     };
 
     PROCESS_INFORMATION processInfo{};
-    LPPROC_THREAD_ATTRIBUTE_LIST attributeList;
-    MakeAttributeList(attributeList);
-    startupInfoEx.lpAttributeList = attributeList;
+
+    if (attributeList)
+    {
```
jsidewhite: I think this if statement should be removed - it's okay to set startupInfoEx.lpAttributeList to nullptr.  It's not okay to leave startupInfoEx.lpAttributeList as stack garbage.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -285,4 +293,34 @@ class PsfPowershellScriptRunner
 
         return true;
     }
+
+    void MakeAttributeList(LPPROC_THREAD_ATTRIBUTE_LIST& attributeList)
+    {
+        SIZE_T AttributeListSize{};
+
+        InitializeProcThreadAttributeList(nullptr, 1, 0, &AttributeListSize);
+        attributeList = (LPPROC_THREAD_ATTRIBUTE_LIST)HeapAlloc(
```
jsidewhite: >attributeList = (LPPROC_THREAD_ATTRIBUTE_LIST)HeapAlloc( [](start = 8, length = 56)

Where is the corresponding delete/free call (DeleteProcThreadAttributeList)?

Can we use an RAII wrapper for this type?

e.g.
    using unique_proc_thread_attribute_list = wil::unique_any<LPPROC_THREAD_ATTRIBUTE_LIST, decltype(&DeleteProcThreadAttributeList), DeleteProcThreadAttributeList>;

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -76,6 +78,7 @@ class PsfPowershellScriptRunner
         bool stopOnScriptError = false;
         std::filesystem::path currentDirectory;
         bool doesScriptExistInConfig = false;
+        LPPROC_THREAD_ATTRIBUTE_LIST attributeList;
```
jsidewhite: Does this leak?  See other comment about RAII.

PsfLauncher/PsfPowershellScriptRunner.h
dhoehna: There is none.  

I've never used the wil::unique_any before.  Where would be the best place to put it?

I'm guessing the destructor of the ScriptInformation struct.  That way the attribute list memory will be freed when the structs go out of scope.

Another idea would be either
1. After the CreateProcess call in StartProcessHelper
2. After the StartProcess calls in PowershellScriptHelper.

Unless I'm wrong, I believe it needs to be called only once because the assignments only copy the pointer, and not the data the pointer points to.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -285,4 +293,34 @@ class PsfPowershellScriptRunner
 
         return true;
     }
+
+    void MakeAttributeList(LPPROC_THREAD_ATTRIBUTE_LIST& attributeList)
```
jsidewhite: >void [](start = 4, length = 4)

Should return the LPPROC_THREAD_ATTRIBUTE_LIST (or RAII equivalent) rather than using an out param.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -65,6 +65,8 @@ class PsfPowershellScriptRunner
     }
 
 private:
+    DWORD createInContainerAttribute = 0x02;
```
jsidewhite: Add comment here that this is really the PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_DISABLE_PROCESS_TREE value.  It's public and documented here,
    [https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute)

Also mention in comment that child and descendant processes of powershell will be bound to the container when this attribute is set.

PsfLauncher/PsfPowershellScriptRunner.h
jsidewhite: I think you should design a little struct (e.g. MyProcThreadAttributeList) at the top of this file that owns the memory for the PROC_THREAD_ATTRIBUTE_LIST, owns the memory for the DWORD (because this memory must outlive the PROC_THREAD_ATTRIBUTE_LIST memory) that specifies creation in the container, and is responsible for calling DeleteProcThreadAttributeList.

I think we should avoid using HeapAlloc() - just use new and have a standard c++ unique_ptr manage that memory (but the smart ptr should be owned by the MyProcThreadAttributeList struct you made above).

Does that all work?

---
In reply to: [394060771](https://github.com/microsoft/MSIX-PackageSupportFramework/pull/136#discussion_r394060771) [](ancestors = 394060771)

dhoehna: I think this should work.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -36,35 +36,94 @@ class PsfPowershellScriptRunner
...
+            return *(attributeList.get());
+        }
+
+        void Initilize()
+        {
+            SIZE_T AttributeListSize{};
+            InitializeProcThreadAttributeList(nullptr, 1, 0, &AttributeListSize);
+
+            attributeList = 
+                std::make_unique<LPPROC_THREAD_ATTRIBUTE_LIST>((LPPROC_THREAD_ATTRIBUTE_LIST) operator new(AttributeListSize));
```
jsidewhite: attributeList = std::unique_ptr<PROC_THREAD_ATTRIBUTE_LIST>(reinterpret_cast<PROC_THREAD_ATTRIBUTE_LIST*>(new BYTE[AttributeListSize]));

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -36,35 +36,94 @@ class PsfPowershellScriptRunner
...
+        // To make sure the attribute value persists we cache it here.
+        // 0x02 is equivalent to PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_DISABLE_PROCESS_TREE
+        // The documentation can be found here:
+        //https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute
+        // This attribute tells PSF to
+        // 1. Create Powershell in the same container as PSF.
+        // 2. Any windows Powershell makes will also be in the same container as PSF.
+        // This means that powershell, and any windows it makes, will have the same restrictions as PSF.
+        DWORD createInContainerAttribute = 0x02;
+        std::unique_ptr<LPPROC_THREAD_ATTRIBUTE_LIST> attributeList;
```
jsidewhite: >LPPROC_THREAD_ATTRIBUTE_LIST [](start = 24, length = 28)

We want a unique_ptr to a struct (not to the pointer to the struct).  Should be,

std::unique_ptr<PROC_THREAD_ATTRIBUTE_LIST> attributeList;

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -36,35 +36,94 @@ class PsfPowershellScriptRunner
...
+                    1,
+                    0,
+                    &AttributeListSize),
+                "Could not initialize the proc thread attribute list.");
+
+            THROW_LAST_ERROR_IF_MSG(
+                !UpdateProcThreadAttribute(
+                    *(attributeList.get()),
+                    0,
+                    ProcThreadAttributeValue(18, FALSE, TRUE, FALSE),
```
jsidewhite: >ProcThreadAttributeValue(18, FALSE, TRUE, FALSE), [](start = 20, length = 49)

Add comment about how 18 represents 'PROC_THREAD_ATTRIBUTE_DESKTOP_APP_POLICY'

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -36,35 +36,94 @@ class PsfPowershellScriptRunner
...
+        std::unique_ptr<LPPROC_THREAD_ATTRIBUTE_LIST> attributeList;
+
+    public:
+
+        LPPROC_THREAD_ATTRIBUTE_LIST get()
+        {
+            return *(attributeList.get());
+        }
+
+        void Initilize()
```
jsidewhite: >Initilize [](start = 13, length = 9)

Typo

jsidewhite: >Initilize [](start = 13, length = 9)

Actually this should be a constructor - it's not valid to have one of these that isn't initialized.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -76,6 +78,7 @@ class PsfPowershellScriptRunner
         bool stopOnScriptError = false;
         std::filesystem::path currentDirectory;
         bool doesScriptExistInConfig = false;
+        LPPROC_THREAD_ATTRIBUTE_LIST attributeList;
     };
 
     ScriptInformation startingScriptInformation;
```
jsidewhite: What if you keep the attributeList in the ScriptInformation object, but have it be optional?

e.g. std::optional<MyProcThreadAttributeList> attributeList;

dhoehna: It should not be optional.  Scripts will always need the attribute list.  I made the attribute list a member variable so I would not a redundant copy of it.

---
In reply to: [396725258](https://github.com/microsoft/MSIX-PackageSupportFramework/pull/136#discussion_r396725258) [](ancestors = 396725258)

PR #137, opened 2020-03-22
Replacement Post350

TimMangan:
This is a replacement PR for the Post350 PR.


dhoehna:
Look at all those conflicts.  I'll be sure to take a look at them.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -56,11 +66,39 @@ class PsfPowershellScriptRunner
 	//RunStartingScript should return an error only if stopOnScriptError is true
 	void RunStartingScript()
 	{
+		LogString("StartingScript commandString", this->startingScriptInformation.commandString.c_str());
+		LogString("StartingScript currentDirectory", this->startingScriptInformation.currentDirectory.c_str());
+		if (this->startingScriptInformation.runInVirtualEnvironment)
```
dhoehna: runInVirtualEnvironment is taken away.  The script will always run in the virtual environment.  The logging and logic for runInVirtualEnvironment can be removed.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -56,11 +66,39 @@ class PsfPowershellScriptRunner
...
+		else
+		{
+			Log("StartingScript waitForScriptToFinish=false");
+		}
 		RunScript(this->startingScriptInformation);
 	}
 
 	void RunEndingScript()
 	{
+		LogString("EndingScript commandString", this->endingScriptInformation.commandString.c_str());
```
dhoehna: Same here with the runInVirtualEnvironment.

tests/scenarios/FileSystemTest/PrivateProfileTest.cpp
```diff
@@ -0,0 +1,132 @@
...
+    auto otherFilePath = g_packageRootPath / L"TestIniFile.ini";
+
+    std::wstring badValue = L"BadValue";
+    std::wstring updatedValue = L"UpdatedValue";
+    int badValueInt = 666;
+
+    clean_redirection_path();
+
+    int   testLen = 2048*sizeof(wchar_t);
+    wchar_t* buffer = (wchar_t *)malloc(testLen);
```
dhoehna: Unless you have a need for this to be a C-style string can you please use an std::wstring?

tests/fixups/TraceFixup/Logging.h
```diff
@@ -142,11 +142,15 @@ inline void LogString(const char* name, const wchar_t* value)
 }
 inline std::string InterpretStringA(const char* value)
 {
-    return value;
+    if (value != NULL)
```
dhoehna: Please add {} even if the statement spans one line.

fixups/FileRedirectionFixup/PathRedirection.h
```diff
@@ -30,8 +30,8 @@ struct path_redirect_info
     bool shouldReadonly = false;
 };
 
-path_redirect_info ShouldRedirect(const char* path, redirect_flags flags);
-path_redirect_info ShouldRedirect(const wchar_t* path, redirect_flags flags);
+path_redirect_info ShouldRedirect(const char* path, redirect_flags flags, DWORD inst=0);
```
dhoehna: nit: add spaces on both sides of the = sign.

fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -274,16 +276,49 @@ void LogString(const char* name, const wchar_t* value)
 }
 void LogString(const wchar_t* name, const char* value)
 {
-    Log(L"%ls=%s\n", name, value);
+    if ((value != NULL && value[1] != 0x0))
```
dhoehna: Please add {}.

fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -274,16 +276,49 @@ void LogString(const char* name, const wchar_t* value)
...
+        Log(L"%ls=%ls", name, (wchar_t*)value);
 }
 void LogString(const wchar_t* name, const wchar_t* value)
 {
-    Log(L"%ls=%ls\n", name, value);
+    if ((value != NULL && ((char*)value)[1] == 0x0))
+        Log(L"%ls=%ls\n", name, value);
+    else
+        Log(L"%ls=%s", name, (char*)value);
+}
```
dhoehna: Please add one blank line after each method.

fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -507,16 +579,24 @@ std::string UrlDecode(std::string str)
             ret += str[i];
         }
         else 
-        {
-            sscanf_s(str.substr(i + 1, 2).c_str(), "%x", &ii);
-            ch = static_cast<char>(ii);
-            ret += ch;
-            i = i + 2;
+        {            
+            if (sscanf_s(str.substr(i + 1, 4).c_str(), "%x", &ii) == 1)
```
dhoehna: What is the following loop supposed to accomplish.  I can't figure it out.

fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -532,54 +612,129 @@ std::wstring UrlDecode(std::wstring str)
...
-    {
-        found = old_string.find("file:/", 0);
-    }
-    if (found != 0)
+    int count = 0;
+    size_t found = old_string.find("file:\\\\",0);
+    if (found == 0)
     {
-        found = old_string.find("FILE:\\", 0);
+        count = 7;
```
dhoehna: Can you please replace the magic number 7 with a constant so we can have a way to tell what this number is for.

fixups/FileRedirectionFixup/PathRedirection.cpp
```diff
@@ -980,19 +1172,19 @@ static path_redirect_info ShouldRedirectImpl(const CharT* path, redirect_flags f
...
-        LogString(L"\t\tFRF Virtualized", vfspath.drive_absolute_path);
+        LogString(inst, L"\t\tFRF Virtualized", vfspath.drive_absolute_path);
     }
 
     // Figure out if this is something we need to redirect
     for (auto& redirectSpec : g_redirectionSpecs)
     {
-        LogString(L"\t\tFRF Check against: base", redirectSpec.base_path.c_str());
+        //LogString(inst, L"\t\tFRF Check against: base", redirectSpec.base_path.c_str());
         if (path_relative_to(vfspath.drive_absolute_path, redirectSpec.base_path))
```
dhoehna: Just as a thought, but I don't like this giant nested if-else blocks.  I don't know how to get rid of it at the moment, but I would like them taken away in the future.

dhoehna: Not saying the if-else chain is anyones fault.  Just voicing my thoughts.

PsfLauncher/main.cpp
```diff
@@ -35,7 +35,8 @@ using namespace std::literals;
 // Forward declarations
 void LogApplicationAndProcessesCollection();
 int launcher_main(PCWSTR args, int cmdShow) noexcept;
-void GetAndLaunchMonitor(const psf::json_object &monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr);
+bool LaunchViaCreateProcessInContainer(const std::wstring app, const std::wstring fullCmd, const std::wstring dir, bool cmdShow);
```
dhoehna: The process will always be launched in the container now.  I use the File Proc Attributes on the powershell exe only.  THis can be removed.

dhoehna:
I see that there are many fixup changes/additions.  That is cool.  With all these fixups in place the Format.csl file located at MSIX-PackageSupportFramework\xmlToJsonConverter needs to be modified to fit with all these fixups that have been coming along.

PsfLauncher/PsfPowershellScriptRunner.h
TimMangan: OK. But leaving the logging in; just like other log entries in the method, they only appear if using the debug build and are helpful if troubleshooting scripts.

PsfLauncher/PsfPowershellScriptRunner.h
TimMangan: Ditto.

tests/scenarios/FileSystemTest/PrivateProfileTest.cpp
TimMangan: This is exactly what a legacy app would do when calling GetPriviate..., isn't that what we want to test?

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: The purpose of the method is to replacing references like "%20" in a string.  This is done in a loop as there may be more than one instance in the input string. You may have a more standard function to do this somewhere, but I didn't find one.

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: Yeah, I hear you. I think this is still short enough to be managable without breaking out more functions.

PsfLauncher/main.cpp
TimMangan: Are you sure? From what I can tell only the powershell launches for SCRIPTS are fixed.  When PsfLauncher is asked to run a non exe file (originally from a shortcut pointing to a file) this is performed using shell launching in this file, and it seems to always be outside of the container.  I will remove the disabled code, but I'm thinking that we end up putting something like it back in place.  Even if we only cover a few file types like cmd/bat/ps1 we will need need some of these to run inside the container, and all of them in the container should be strived for.

PsfLauncher/PsfPowershellScriptRunner.h
TimMangan: By the way I am assuming you meant running in the VE is fixed.  The json still provides a selectable option on this, right?  Sometimes we NEED the script to run external; most times internal.

tests/scenarios/FileSystemTest/PrivateProfileTest.cpp
TimMangan: Addressed in Commit 1bdc01a3

tests/fixups/TraceFixup/Logging.h
TimMangan: Addressed in Commit 1bdc01a3

fixups/FileRedirectionFixup/PathRedirection.h
TimMangan: Addressed in Commit 1bdc01a3

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: Addressed in Commit 1bdc01a3

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: Addressed in Commit 1bdc01a3

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: Addressed in Commit 1bdc01a3

PsfLauncher/main.cpp
TimMangan: Addressed in Commit 1bdc01a3, although my previous comment still is in play.

TimMangan:
Hopefully I have addressed what is needed for this PR?  My assumption was the xmlToJsonConverter is a seperate task for you.

PsfLauncher/PsfPowershellScriptRunner.h
dhoehna: Hey Tim.

Leaving the debugger code in is fine.

I'm sorry but it was a decision made higher up.  Scripts are not allowed ever to run outside of the environment.  Letting scripts run outside of the environment breaks the MSIX promise that everything is contained.  There is no option anymore to have scripts run outside the VE.  Sorry about that.

tests/scenarios/FileSystemTest/PrivateProfileTest.cpp
dhoehna: I see your point.  If you want you can change it back to a C-Style string.  Or, if you want to go above and beyond, you can test both.

tests/fixups/TraceFixup/Logging.h
dhoehna: Thank you.

fixups/FileRedirectionFixup/PathRedirection.cpp
dhoehna: I noticed that there are two places where this logic is used.

For now can you please add a comment explaining what this does?  If it turns out we need to add this logic in again somewhere let's move it into a method.

fixups/FileRedirectionFixup/PathRedirection.cpp
dhoehna: Coolio.

PsfLauncher/main.cpp
dhoehna: I see.  I'm sorry, I was confused.  You are correct.  PSF launches only scripts and powershell.exe in a container, nothing else.  

The method LaunchViaCreateProcessInContainer is for the main application that PSF will launch, right?

dhoehna:
At the moment, since I didn't ask, I can modify the converted.  Next time though, the xsl file will need to be updated to support any fixup modifications.

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: Done.

PsfLauncher/PsfPowershellScriptRunner.h
TimMangan: OK. Doesn't accomplish what they think it does and will make more resistance to IT Pros converting to MSIX.  But, that's not my problem either. Needs to be communicated out, though.

fixups/FileRedirectionFixup/PathRedirection.cpp
TimMangan: Noticed that I missed the wstring version on this.  Updated both string and wstring versions, possibly with some simpler code.

dhoehna:
Everything looks good.  Is it okay if I edit your branch to fix the merge conflicts?

TimMangan:
> Everything looks good.  Is it okay if I edit your branch to fix the merge conflicts?

Yes.  Let me know if I need to give you permission or something like that.

dhoehna:
Fixed the merge conflicts.  Before I complete the PR I'm going to make sure your code will build and test.

PR #150, opened 2020-07-08
Add new RegLegacyFixups fixup type "FakeDelete"

TimMangan:
Added a second type of fixup for RegLegacyFixups.

This type allows specification of registry keys/values for situations where the application attempts to cleanup registry entries that are part of the package.  As MSIX does not allow this, by faking the deletion and returning a success code to the application it can shut down.  Of course, there is no guarantee that this will solve the problem for the application, but in situations where the app is just doing a little cleanup on shutdown it can prevent application crashes on shutdown.

A great example of such an application is Fiddler (which also needs the ModifyKeyAccess fixup type).


fixups/RegLegacyFixups/FunctionImplementations.h
```diff
@@ -197,6 +222,9 @@ namespace impl
     //inline auto NtQueryInformationFile = WINTERNL_FUNCTION(winternl::NtQueryInformationFile);
     //inline auto NtQueryValueKey = WINTERNL_FUNCTION(winternl::NtQueryValueKey);
 
+
+   // inline auto RegDeleteKey = WINTERNL_FUNCTION(winternl::RegDeleteKey);
```
dhoehna, line 226: Please remove commented out code.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -210,24 +205,52 @@ void InitializeConfiguration()
...
+                                Log(L"Pattern:        %Ls\n", patternString.data());
+                                recordItem.fakeDeleteKey.patterns.push_back(patternString.data());
+                            }
+                            Log("RegLegacyFixups:      have patterns\n");
 
+                            specItem.remediationRecords.push_back(recordItem);
+                        }
+                        else
+                        {
+                            ///specItem.remeditaionType = Reg_Remediation_Type_Unknown;
```
dhoehna, line 249: Please remove commented out code.

fixups/RegLegacyFixups/FunctionImplementations.h
TimMangan: Done.

fixups/RegLegacyFixups/InitializeFixup.cpp
TimMangan: Done.

PR #173, opened 2021-02-18
Fix FRF Redundancy Check

TimMangan:
Fixing check on the return value when looking to see if redirection should not be offered because the input path is already redirected.

### DIFF
--- a/fixups/DynamicLibraryFixup/FunctionImplementations.h
+++ b/fixups/DynamicLibraryFixup/FunctionImplementations.h
@@ -31,6 +31,13 @@ inline std::wstring InterpretStringW(const wchar_t* value)
     return value;
 }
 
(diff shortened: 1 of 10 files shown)


fixups/DynamicLibraryFixup/FunctionImplementations.h
```diff
@@ -31,6 +31,13 @@ inline std::wstring InterpretStringW(const wchar_t* value)
     return value;
 }
 
+inline std::wstring GetFilenameOnly(std::wstring path)
+{
+    size_t index = path.find_last_of(L"\\", std::wstring::npos);
+    if (index == std::wstring::npos)
```
dhoehna, line 37: Brackets please.

TimMangan:
The only file I changed is PathRedirection.cpp I will close this PR and build a new fork, yet again, to prevent this crap.

Issue #167, opened 2021-01-12
PsfRuntime CreateProcessHook issue with mixed bitness


TimMangan:
When a processes under the influence of the PSF starts a new process, this action is caught by the version of PsfRuntime in the calling process (CreateProcessHook).  This code will create the new process as suspended, and then (attempt to) inject a fresh copy of PsfRuntime into the new process.

The current code (around lines 164-173 of the develop branch) uses psf::runtime_dll_name which returns the name assuming the bitness of the calling process.  So if a 32-bit process is starting a 64-bit process that injection will fail (or visa versa).  The result is that the new process never gets a chance to run.

This code will need to be updated. Probably best is to look at the bitness of the started process to select the correct name to use, but if not perhaps it can at least retry a different bitness upon error.  PsfRundll.exe is also a possibility to consider, but I think that you'd still need to modify code to look at the bitness of the started process.

npuvvada:
This fix is available from [PSF release : version 1.0.220926.1](https://github.com/microsoft/MSIX-PackageSupportFramework/releases/tag/1.0.220926.1)

PR #176, opened 2021-02-26
Develop

TimMangan:
Fix for Issue #167 
Also cleanup on process launch attribute check and better debug logging on process launching.

### DIFF
--- a/PsfLauncher/main.cpp
+++ b/PsfLauncher/main.cpp
@@ -141,15 +141,15 @@ int launcher_main(PCWSTR args, int cmdShow) noexcept try
         {
             fullargs = args;
         }
-
-        LogString("Process Launch: ", fullargs.data());
+        LogString("Process Launch: ", exePath.c_str());
+        LogString("     Arguments: ", fullargs.data());
         LogString("Working Directory: ", currentDirectory.c_str());
 
         //THROW_IF_FAILED(StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE));
         HRESULT hr = StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE);
         if (hr != ERROR_SUCCESS)
         {
-            Log("Error return from launching process.");
+            Log("Error return from launching process 0x%x.", GetLastError());
         }
     }
     else
--- a/PsfRuntime/CreateProcessHook.cpp
+++ b/PsfRuntime/CreateProcessHook.cpp
@@ -29,6 +29,66 @@ using namespace std::literals;
 
 void Log(const char* fmt, ...);
 
+// Function to determine if this process should get 32 or 64bit dll injections
+// Returns 32 or 64 (or 0 for error)
+typedef BOOL(WINAPI* LPFN_ISWOW64PROCESS2) (HANDLE, PUSHORT, PUSHORT);
+USHORT ProcessBitness(HANDLE hProcess)
+{
+    USHORT pProcessMachine;
+    USHORT pMachineNative;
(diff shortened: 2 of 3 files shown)


PsfRuntime/CreateProcessHook.cpp
```diff
@@ -156,77 +216,106 @@ BOOL WINAPI CreateProcessFixup(
...
         {
             // Possibly the dll is in the folder with the exe and not at the package root.
-            Log("\t%s not found at package root, try target folder.", targetDll);
+            Log("\t%ls not found at package root, try target folder.", wtargetDllName.c_str());
 
             std::filesystem::path altPathToExeRuntime = exePath.data();
-            static const auto altPathToPsfRuntime = (altPathToExeRuntime.parent_path() / psf::runtime_dll_name).string();
-            targetDll = altPathToPsfRuntime.c_str();
+            static const auto altPathToPsfRuntime = (altPathToExeRuntime.parent_path() / pathToPsfRuntime.c_str()).string();
+            //targetDll = altPathToPsfRuntime.c_str();
```
dhoehna, line 249: NIT: Remove the commented out line.

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -156,77 +216,106 @@ BOOL WINAPI CreateProcessFixup(
...
-                ::TerminateProcess(processInformation->hProcess, ~0u);
-                ::CloseHandle(processInformation->hProcess);
-                ::CloseHandle(processInformation->hThread);
-
-                ::SetLastError(err);
-                return FALSE;
+                Log("\t%ls unable to inject, skipping.", targetDllPath);
+                // We failed to detour the created process. Assume that it the failure was due to an architecture mis-match
+                // and try the launch using PsfRunDll
+                //if (!::DetourProcessViaHelperDllsW(processInformation->dwProcessId, 1, &targetDllPath, CreateProcessWithPsfRunDll))
```
dhoehna, line 295: Remove this block of commented out code.

PsfRuntime/CreateProcessHook.cpp
TimMangan: Done.

PsfRuntime/CreateProcessHook.cpp
TimMangan: Done.

PsfLauncher/main.cpp
```diff
@@ -141,15 +141,15 @@ int launcher_main(PCWSTR args, int cmdShow) noexcept try
...
+        LogString("Process Launch: ", exePath.c_str());
+        LogString("     Arguments: ", fullargs.data());
         LogString("Working Directory: ", currentDirectory.c_str());
 
         //THROW_IF_FAILED(StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE));
         HRESULT hr = StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE);
         if (hr != ERROR_SUCCESS)
         {
-            Log("Error return from launching process.");
+            Log("Error return from launching process 0x%x.", GetLastError());
```
dhoehna, line 152: GetLastError returns a decimal, not a hex.  Please remove the 0x

TimMangan: I am going to disagree with the comment.  You can insist if you want, but...

The function returns an binary value, which is displayable as a decimal or hex form.  Most computer professionals, the only ones to be looking at the debug console port, would prefer to see this in hex form as this is the most traditional format for error codes.  This makes a return code of 0x70003 be familiar as a file/registry not found even though the error is more specific.

dhoehna: Interesting.  Thank you for your input.  

I prefer to see decimal.  I wonder if I'm the minority.

Either way, keep the change.  If people start asking for decimal it can be changed.

Issue #179, opened 2021-05-14
Issues with PowerShell start and end scripts


Skinheed:
Hi,

I am seeing a problem with the PSF launcher where it uses the working directory of an MSIX shortcut to locate the StartingScriptWrapper.ps1 file. I have had this confirmed by Tim Mangan and Darren Hoehna. I am seeing the problem when I use PSF Tooling and Advanced Installer Express.

TimMangan:
A potential fix for this issue has been pulled into my fork as noted above.

npuvvada:
This fix is available from [PSF release : version 1.0.220926.1](https://github.com/microsoft/MSIX-PackageSupportFramework/releases/tag/1.0.220926.1)

PR #183, opened 2021-07-17
Fix for issue #179 Not finding StartingScriptWrapper.ps1

TimMangan:
Ensure that if StartingScriptWrapper.ps1 is not in the working directory that it is found as long as it is somewhere inside the package.


dhoehna:
Thank you.

PR #188, opened 2021-12-07
Ravishroshan/platform upgrade

ravishroshanms:
Updated the projects to be able to be built on Visual Studio 2019 without having to depend on Visual studio 2017 build tools and SDKs.

### DIFF
--- a/fixups/EnvVarFixup/EnvVarFixup.xml
+++ b/fixups/EnvVarFixup/EnvVarFixup.xml
@@ -3,7 +3,7 @@
          SchemaLocation="https://github.com/Microsoft/MSIX-PackageSupportFramework/blob/master/MetadataSchema.xml ..\MetadataSchema.xsd">
   >
   <Version>1.0.0</Version>
-  <MinimumWindowsPlatform>10.0</MinimumWindowsPlatform>
+  <MinimumWindowsPlatform>10.0.14393.0</MinimumWindowsPlatform>
   <Description>
     To redirect environment variable calls to work within the package.
   </Description>
--- a/fixups/EnvVarFixup/EnvVarFixup.xml
+++ b/fixups/EnvVarFixup/EnvVarFixup.xml
@@ -3,7 +3,7 @@
          SchemaLocation="https://github.com/Microsoft/MSIX-PackageSupportFramework/blob/master/MetadataSchema.xml ..\MetadataSchema.xsd">
   >
   <Version>1.0.0</Version>
-  <MinimumWindowsPlatform>10.0.14393.0</MinimumWindowsPlatform>
+  <MinimumWindowsPlatform>10.0</MinimumWindowsPlatform>
   <Description>
     To redirect environment variable calls to work within the package.
   </Description>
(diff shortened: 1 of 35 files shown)


fixups/EnvVarFixup/EnvVarFixup.xml
```diff
@@ -3,7 +3,7 @@
          SchemaLocation="https://github.com/Microsoft/MSIX-PackageSupportFramework/blob/master/MetadataSchema.xml ..\MetadataSchema.xsd">
   >
   <Version>1.0.0</Version>
-  <MinimumWindowsPlatform>10.0.14393.0</MinimumWindowsPlatform>
+  <MinimumWindowsPlatform>10.0</MinimumWindowsPlatform>
```
dhoehna, line 6: This should stay at 14393.0.  PSF guarantees an RS1+ compatibility (Current it is RS2+).  Moving this higher would break that promise.

dhoehna: Same for every version change.

fixups/RegLegacyFixups/RegLegacyFixups.vcxproj
```diff
@@ -1,222 +1,222 @@
...
-  <ItemGroup>
-    <None Include="readme.md" />
-  </ItemGroup>
-  <ItemGroup>
-    <Xml Include="RegLegacyFixupsMetadata.xml" />
-  </ItemGroup>
-  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-  <ImportGroup Label="ExtensionTargets">
-  </ImportGroup>
+<?xml version="1.0" encoding="utf-8"?>
```
dhoehna, line 1: Any idea why the whole vcxproj file is diffed?

include/psf_config.h
```diff
@@ -53,6 +53,7 @@ namespace psf
 
     struct json_value
     {
+        virtual ~json_value() {}
```
dhoehna, line 56: What is the destructor for?  PSF worked without it.

dhoehna: I would say to not include a destructor until it has custom logic.

fixups/RegLegacyFixups/RegLegacyFixups.vcxproj
ravishroshanms: This was done due to some indentation change which was automatically done by VS

fixups/EnvVarFixup/EnvVarFixup.xml
ravishroshanms: Reverted this change

include/psf_config.h
ravishroshanms: The new version of the windows SDK is throwing warning in case where the destructor is missing. Moreover it make sense to have the virtual destructor here and we are mandated to have a definition attached with the virtual destructor as there are never going to be overridden.

BenJKuhn: base class needs to have a virtual destructor for derived to clean up properly if delete is called on a base pointer. This error is likely from the toolset change. Either way this is goodness.

BenJKuhn (approved):

kyelamarthi1 (approved):
LGTM .

dhoehna:
In light of PSF getting off the ground.  I'm going to merge this PR without the build being successful.  Ravish is working on getting the build green.

Keep in mind that I will block any other PR's until the pipeline is green.

PR #191, opened 2022-02-01
Added the test ini file in the redirection pattern to fix failing test

ravishroshanms:
Private profile tests were failing because the file used in the test for updating the string was not being identified from the redirection pattern.

> Added the file name in the pattern


Vipin-Dalal (approved):
Added the missing ".ini" files for redirection

PR #192, opened 2022-02-16
Fixed failing architecture tests

ravishroshanms:
Detour fails due to an architecture mis-match. Trying the launch using PsfRunDll.


dhoehna:
PSF should adopt RAII for handles and such.

Is this PR to close the resources?

ravishroshanms:
We are using DetourProcessViaHelperDllsW API to launch the process using PsfRunDll when DetourUpdateProcessWithDll fails due to architecture mismatch. This change fixes the architecture tests which are failing in the develop currently. I agree that we should have RAII for handles, but right now I have handled them as they were being handled till now. Will take that up separately later.

PR #197, opened 2022-07-12
Fix Powershell script fail scenarios

npuvvada:
1. PowershellScript test was failing for three scenarios. 
PSOnlyStart, PSBothStartingFirst and PSScriptWithArg scenarios are failing.

Proper command line format (with appropriate quotation marks) is used to fix the issue. 

2. Powershell script and appplication fails to start when "WaitForScriptToFinish" parameter is set to False in config.

call detach on thread where power shell is run and added test case to ensure both application and power shell are running parallel


npuvvada:
moved to wrong target branch. closing it

PR #199, opened 2022-07-26
Handling script arguments with space

npuvvada:
* Updated README.md to give script Arguments with space in config.json
* Test case to handle script arguments with space is updated. (PowershellScriptTest_PSScriptWithArg)


rahul89dadhich (approved):
looks good

PR #211, opened 2022-09-26
PSFMontor folder restructuring in Release folder

npuvvada:
-Include x86 DLLs alone in psfMonitor x86 folder
-Use x86 folder name instead of Win32 in psfMonitor


npuvvada:
Close pull request as develop to master pull request was raised by mistake

PR #218, opened 2022-11-02
Added fix for consistent prefix matching for HKLM and HKCU (#217)

npuvvada:
* Added fix for consistent prefix matching for HKLM and HKCU

* Removed duplicate code


rahul89dadhich (approved):
looks good

PR #220, opened 2022-12-12
CreateProcess Argument redirection feature

npuvvada:
When an external (to the container) process is launched with arguments including appData path, these are not virtualized and in turn application tries to access files in native app data path instead of per user per app data path because of which it fails to find it.

## What changed?
This feature handles redirecting createProcess argument calls from native app data to per user per app data when corresponding files are found in per user per appdata file.

## How was the change tested?
-A test application is added which triggers cmd.exe(C:\\Windows\\System32\\cmd.exe) to rename an existing file in appdata 
 folder (passed as a parameter). Renaming would be successful when Argument is redirected to per user per appdata folder.


PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+}
+
+template <typename CharT>
+bool is_path_relative(const CharT* path, const std::filesystem::path& basePath)
+{
+    return std::equal(basePath.native().begin(), basePath.native().end(), path, psf::path_compare{});
+}
+
+template <typename CharT>
+bool IsUnderUserAppDataRoamingandReplace(const CharT* fileName, CharT* cnvtCmdLine)
```
rahul89dadhich, line 56: Can we break this to 2 method? 1 just to check and other to replace?

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+            strcatImpl(cnvtCmdLine, remBufSize, fileName + (roamingAppData.native().length()));
+        }
+    }
+    return result;
+}
+
+template <typename CharT>
+bool IsUnderUserAppDataLocalandReplace(const CharT* fileName, CharT* cnvtCmdLine)
+{
+    bool isUnderAppDataLocal = false;
```
rahul89dadhich, line 102: I think if you extract in 2 method, duplicate code can be removed

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -153,9 +155,17 @@ BOOL WINAPI CreateProcessFixup(
         processInformation = &pi;
     }
 
+    // Redirect createProcess arguments if any in native app data to per user per app data
+    CharT* cnvtCmdLine = new CharT[MAX_CMDLINE_PATH];
+    if (cnvtCmdLine)
```
rahul89dadhich, line 160: if this is command line input, it can't exceed 8191, why reserving extra? else changes the define string

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -153,9 +155,17 @@ BOOL WINAPI CreateProcessFixup(
         processInformation = &pi;
     }
 
+    // Redirect createProcess arguments if any in native app data to per user per app data
+    CharT* cnvtCmdLine = new CharT[MAX_CMDLINE_PATH];
+    if (cnvtCmdLine)
+    {
+        cnvtCmdLine[0] = (CharT)nullptr;
```
rahul89dadhich, line 162: Can we use memcpy to initialize complete array?

PsfRuntime/readme.md
```diff
@@ -57,3 +57,7 @@ The PSF Runtime then calls `PSFInitialize` within a Detours transaction, failing
...
+
+## Argument Redirection
+This feature helps to redirect any createProcess call arguments from native appdata to per user per app data when the target file provided in arguments is actually present in per user per app data folder. When any child process thats not part of the container is triggered, detoured createProcess will check each argument passed and will convert it to corresponding path in per user per appdata folder.
```
rahul89dadhich, line 62: not part of the container -> not running in package context

rahul89dadhich (changes requested):
Please check comments

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+
+#define MAX_CMDLINE_PATH 32767
```
dhoehna, line 7: Is this constant not defined in a windows header?

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+auto GetFileAttributesImpl = psf::detoured_string_function(&::GetFileAttributesA, &::GetFileAttributesW);
+
+template <typename CharT>
+errno_t strcatImpl(CharT* dest, rsize_t destBufSize, CharT const* src)
+{
+    if (std::is_same<CharT, char>::value)
+    {
+        return strcat_s((char*)dest, destBufSize, (const char*)src);
+    }
+    else if (std::is_same<CharT, wchar_t>::value)
```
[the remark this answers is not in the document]
dhoehna: Or remove `if`.  People know we're dealing only with char and wchar.

PsfRuntime/ArgRedirection.h
dhoehna, line 56: Please type out cnvt.  Cmd is known colloquially no need to change that.

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+    return std::equal(basePath.native().begin(), basePath.native().end(), path, psf::path_compare{});
+}
+
+template <typename CharT>
+bool IsUnderUserAppDataRoamingandReplace(const CharT* fileName, CharT* cnvtCmdLine)
+{
+    bool isUnderAppDataRoaming = false;
+    bool result = false;
+
+    if (fileName == NULL)
```
dhoehna, line 61: nullptr

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+    {
+        return result;
+    }
+    constexpr wchar_t root_local_device_prefix[] = LR"(\\?\)";
+    constexpr wchar_t root_local_device_prefix_dot[] = LR"(\\.\)";
+
+    if (std::equal(root_local_device_prefix, root_local_device_prefix + 4, fileName))
```
dhoehna, line 68: Magic number.  Please make a constant.

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+    else if (std::equal(root_local_device_prefix_dot, root_local_device_prefix_dot + 4, fileName))
+    {
+        isUnderAppDataRoaming = is_path_relative(fileName + 4, psf::known_folder(FOLDERID_RoamingAppData));
+    }
+    else
+    {
+        isUnderAppDataRoaming = is_path_relative(fileName, psf::known_folder(FOLDERID_RoamingAppData));
+    }
+
+    size_t remBufSize = MAX_CMDLINE_PATH - strlenImpl(cnvtCmdLine);
```
dhoehna, line 81: What is rem?

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+            result = true;
+            strcatImpl(cnvtCmdLine, remBufSize, (CharT*)(packageLocalAppDataPath.c_str()));
+            strcatImpl(cnvtCmdLine, remBufSize, fileName + (localAppdata.native().length()));
+        }
+    }
+    return result;
+}
+
+template <typename CharT>
+void convertCmdLineParameters(CharT* inpCmdLine, CharT* cnvtCmdLine)
```
dhoehna, line 144: What is inp?

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+{
+    CharT* next_token;
+    size_t cmdLinelen = strlenImpl(inpCmdLine);
+    CharT* cmdLine = new CharT[cmdLinelen+1];
```
dhoehna, line 148: Please wrap this into a std::unique_ptr<CharT>.  This avoids us having to manually delete the array,

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+    if (cmdLine)
```
dhoehna, line 149: I don't understand if(cmdLine).  This should always be true because you just allocated memory, right?

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,186 @@
...
+            strcatImpl(cnvtCmdLine, remBufSize, token);
+        }
+        token = strtokImpl((CharT*)nullptr, (CharT*)L" ", &next_token);
+        if (token != nullptr)
+        {
+            remBufSize = MAX_CMDLINE_PATH - strlenImpl(cnvtCmdLine);
+            strcatImpl(cnvtCmdLine, remBufSize, (CharT*)L" ");
+        }
+    }
+    delete[] cmdLine;
```
dhoehna, line 184: Please remove and use std::unique_ptr

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -153,9 +155,17 @@ BOOL WINAPI CreateProcessFixup(
         processInformation = &pi;
     }
 
+    // Redirect createProcess arguments if any in native app data to per user per app data
+    CharT* cnvtCmdLine = new CharT[MAX_CMDLINE_PATH];
```
dhoehna, line 159: Please use std::unique_ptr

PsfRuntime/ArgRedirection.h
npuvvada: remaining buffer size

PsfRuntime/ArgRedirection.h
npuvvada: Removed duplicate functions for appdatalocal and appdataroaming

PsfRuntime/ArgRedirection.h
npuvvada: inputCmdLine

PsfRuntime/ArgRedirection.h
npuvvada: This is to avoid static analyzer warnings

PsfRuntime/CreateProcessHook.cpp
npuvvada: This is commandLine argument in createProcess win API call whose maximum lenggth can be 32,767 characters
https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa

PsfRuntime/ArgRedirection.h
dhoehna: Please use constexpr.

PsfRuntime/ArgRedirection.h
dhoehna: Please change inp to input.

include/utilities.h
```diff
@@ -223,3 +224,48 @@ inline wide_argument_string widen_argument(const wchar_t* str) noexcept
 {
     return wide_argument_string{ str };
 }
+
+template <typename CharT>
+errno_t strcatImpl(CharT* dest, rsize_t destBufSize, CharT const* src)
+{
+    if (std::is_same<CharT, char>::value)
+    {
+        return strcat_s((char*)dest, destBufSize, (const char*)src);
```
dhoehna, line 233: Do not use C-Style casts.

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -165,8 +175,12 @@ BOOL WINAPI CreateProcessFixup(
         startupInfo,
         processInformation))
     {
+        if (cnvtCmdLine != nullptr)
```
dhoehna: Always use {} even if the block is 1 line.

However, if you use std::unique_ptr you don't have to wrry about manually deleting the array.

PsfRuntime/ArgRedirection.h
dhoehna: What is the warning?

npuvvada: nullptr deference warning. 
Many static code analyzers can detect when this check (for nullptr after dynamically allocating memory) is missing and issue a warning.

include/utilities.h
```diff
@@ -223,3 +224,48 @@ inline wide_argument_string widen_argument(const wchar_t* str) noexcept
...
+template <typename CharT>
+CharT* strtokImpl(CharT* inpStr, CharT const* delim, CharT** token)
+{
+    if (std::is_same<CharT, char>::value)
+    {
+        return (CharT*)(strtok_s((char*)inpStr, (const char*)delim, (char**)token));
+    }
+    else
+    {
+        return (CharT*)(wcstok_s((wchar_t*)inpStr, (const wchar_t*)delim, (wchar_t**)token));
```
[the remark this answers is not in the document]
npuvvada: strtok/wcstok split a string into tokens (substrings) based on a specified delimiter. Using it here to check each command line argument. 

each argument is expected to be wchar_t* type for wcstok_s. typecasting is done after confirming that template type(CharT) is actually of type wchar_t. Hence, this casting is safe.

include/utilities.h
npuvvada: Here c-style casting is done only after confirming the casting is safer to be done. CharT was already confirmed to be of type char* and then casting is done. So, I thought c++ style casting would not give an advantage over c-style in this case.

dhoehna: I see and understand you stance.  However, C++ style casting has the benefit of telling the reader what kind of cast is happening.  Also, where is the check to see if the cast is safe?  If I can't find it, other people can't either.

npuvvada: Agree, I will change it to c++ style casting. 
By confirming the casting is safer, i meant template parameter of the parameters used in strcat_s(CharT) is checked before casting.

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,110 @@
...
+        return result;
+    }
+    constexpr wchar_t root_local_device_prefix[] = LR"(\\?\)";
+    constexpr wchar_t root_local_device_prefix_dot[] = LR"(\\.\)";
+
+    GUID appDataFolderId = AppDataLocal ? FOLDERID_LocalAppData : FOLDERID_RoamingAppData;
+
+    if (std::equal(root_local_device_prefix, root_local_device_prefix + wcslen(root_local_device_prefix), fileName))
+    {
+        isUnderAppData = is_path_relative(fileName + wcslen(root_local_device_prefix), psf::known_folder(appDataFolderId));
```
ravishroshanms, line 27: Do we have a test case for this?

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,110 @@
...
+    }
+    else if (std::equal(root_local_device_prefix_dot, root_local_device_prefix_dot + wcslen(root_local_device_prefix_dot), fileName))
+    {
+        isUnderAppData = is_path_relative(fileName + wcslen(root_local_device_prefix_dot), psf::known_folder(appDataFolderId));
```
ravishroshanms, line 31: Same here. Do we have a test case?

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -24,6 +24,8 @@
 #include <psf_framework.h>
 
 #include "Config.h"
+#include <known_folders.h>
```
ravishroshanms, line 27: Is this even needed?

PsfRuntime/PsfRuntime.vcxproj.filters
```diff
@@ -29,5 +29,8 @@
     <ClInclude Include="JsonConfig.h">
       <Filter>inc</Filter>
     </ClInclude>
+    <ClInclude Include="ArgRedirection.h">
+      <Filter>inc</Filter>
+    </ClInclude>
   </ItemGroup>
-</Project>
```
ravishroshanms, line 33: Please correct end of line

PsfRuntime/readme.md
```diff
@@ -57,3 +57,9 @@ The PSF Runtime then calls `PSFInitialize` within a Detours transaction, failing
 
 ## Runtime Requirements
 As a part of its initialization, the PSF Runtime queries information about its environment that it then caches for later use. A few examples include parsing the `config.json`, caching the path to the package root, and caching the package name, among a couple other things. If any of these steps fail, e.g. because something is not present/cannot be found or any other failure, then the PSF Runtime dll will fail to load, which likely means that the process fails to start. Note that this implies the requirement that the application be running with package identity. There have been past conversations on adding support for a "debug" mode that works around this restriction (e.g. by using a fake package name, executable directory as the package root, etc.), but its benefit is questionable and has not yet been implemented.
+
+## Argument Redirection
```
ravishroshanms, line 61: This content looks more appropriate for Release notes and not for Readme. This redirection is going to happen internally without user having any control over it. In case if we still decide to have it here, we would need to reword it.

include/utilities.h
```diff
@@ -223,3 +224,48 @@ inline wide_argument_string widen_argument(const wchar_t* str) noexcept
...
+    {
+        return wcslen((const wchar_t*)inpStr);
+    }
+}
+
+template <typename CharT>
+bool is_path_relative(const CharT* path, const std::filesystem::path& basePath)
+{
+    return std::equal(basePath.native().begin(), basePath.native().end(), path, psf::path_compare{});
+}
```
ravishroshanms, line 271: End of line

ravishroshanms (changes requested):

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,110 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+
+constexpr int MAX_CMDLINE_PATH = 32767;
```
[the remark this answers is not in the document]
npuvvada: This is chosen as per the maximum length of cmdLineArgument allowed as per createProcess documentation

https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa
[in, out, optional] lpCommandLine

The command line to be executed.

The maximum length of this string is 32,767 characters, including the Unicode terminating null character. If lpApplicationName is NULL, the module name portion of lpCommandLine is limited to MAX_PATH characters.

PsfRuntime/ArgRedirection.h
```diff
@@ -0,0 +1,110 @@
...
+                size_t bytes_written;
+                std::unique_ptr<char[]> packageAppDataPath_char(new char[packageAppDataPathLen + 1]);
+                wcstombs_s(&bytes_written, packageAppDataPath_char.get(), packageAppDataPathLen + 1, packageAppDataPath.c_str(), packageAppDataPathLen);
+                strcatImpl(cmdLine, remBufSize, (const CharT*)(packageAppDataPath_char.get()));
+            }
+            else
+            {
+                strcatImpl(cmdLine, remBufSize, (CharT*)(packageAppDataPath.c_str()));
+            }
+            strcatImpl(cmdLine, remBufSize, fileName + (appDataPath.native().length()));
```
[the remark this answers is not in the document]
npuvvada: fullPath is wstring. We need to consider if inputCommandLine is char/wchar and then append corresponding version of converted CommandLine

PsfRuntime/ArgRedirection.h
npuvvada: No, but similar piece of code for checking local device/root local device path type is already used in File Redirection fixup and tested to be working fine. We are re-using the same here

PsfRuntime/ArgRedirection.h
npuvvada: No, but similar piece of code for checking local device/root local device path type is already used in File Redirection fixup and tested to be working fine. We are re-using the same here

PsfRuntime/CreateProcessHook.cpp
npuvvada: This is included for calculating appData folder path in argument redirection. Moving it to ArgumentRedirection.h file

PsfRuntime/ArgRedirection.h
ravishroshanms: Sure. Thanks

ravishroshanms (approved):

rahul89dadhich (approved):

PR #221, opened 2022-12-20
New feature to run external processes in package context

npuvvada:
A new field "InPackageContext" is introduced to run spanned process by the external in the current package context. When set to true, all the processes in the process tree spanned by the executable will run in the same package context.


PsfRuntime/CreateProcessHook.cpp
```diff
@@ -153,19 +157,68 @@ BOOL WINAPI CreateProcessFixup(
...
+    auto preventBreakAwayPtr = appConfig->try_get("InPackageContext");
+    if (preventBreakAwayPtr)
+    {
+        preventBreakAway = preventBreakAwayPtr->as_boolean().get();
+    }
+
+    if (preventBreakAway)
     {
-        return FALSE;
+        STARTUPINFOEX startupInfoEx =
```
dhoehna, line 170: StartupInfoEx and the call to CreateProcessImpl can be moved outside of the if block.

Make StartupInfoEx
if(Prevent Breakaway)
    Add breakaway
CreateProcessImpl

include/proc_helper.h
```diff
@@ -0,0 +1,69 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#pragma once
+#include <wil\result.h>
+
+// Define _PROC_THREAD_ATTRIBUTE_LIST as an empty struct because it's internal-only and variable-sized
+struct _PROC_THREAD_ATTRIBUTE_LIST {};
+
+struct MyProcThreadAttributeList
```
dhoehna, line 12: NIT: Change this name to something more descriptive.  Or, at least get rid of "my"

dhoehna:
Awesome job.  I remember this was an issue when I was in India.  Great job figuring out a solution.

PsfRuntime/CreateProcessHook.cpp
npuvvada: you mean to use extended startupInfo even when prevent breakaway is not enabled ?

dhoehna: Yes.  Then you can move the StartupInfoEx outside of the if block.

dhoehna, line 170: Just realized that if preventBreakaway is true this code path ignores StartupInfo.  The data in StartupInfo should be merged with StartupInfoEx.

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -153,19 +157,68 @@ BOOL WINAPI CreateProcessFixup(
...
-        processInformation))
+    auto appConfig = PSFQueryCurrentAppLaunchConfig(true);
+    bool preventBreakAway = false;
+    auto preventBreakAwayPtr = appConfig->try_get("InPackageContext");
+    if (preventBreakAwayPtr)
+    {
+        preventBreakAway = preventBreakAwayPtr->as_boolean().get();
+    }
+
+    if (preventBreakAway)
```
dhoehna, line 168: I'm a bit wishy-washy on "preventBreakaway" because preventBreakaway isn't the action we're taking.  This is to launch the process in the container.

NIT: rename to createProcessesInContainer.

@rahul89dadhich what do you think?

rahul89dadhich: agree, how about "createProcessInAppContext" ?

PsfRuntime/CreateProcessHook.cpp
npuvvada: Oh, Yes. Thanks for pointing it out

PsfLauncher/Readme.md
```diff
@@ -332,6 +335,13 @@ The PSF Launcher supports the use of two special purpose "pseudo-variables". The
 | %MsixPackageRoot% | The root folder of the package. While nominally this would be a subfolder under "C:\\Program Files\\WindowsApps" it is possible for the volume to be mounted in other locations. |
 | %MsixWritablePackageRoot% | The package specific redirection location for this user when the FileRedirectionFixup is in use. | 
 
+### InPackageContext
+This configuration allows all processes created when the executable is run to operate in the same package context. This includes any child processes that might not be part of the package. This might be useful in situations where applications span processes that are outside package but needs to be run in the same package context, or when PSF fixups need to be applied to processes external to the package.
```
rahul89dadhich, line 339: Can we change this description? 
This configuration allows all processes created when the executable is run to operate in the same package context. -->
This Configuration allows all created process from a running executable to operate in same package context.?

PsfRuntime/CreateProcessHook.cpp
```diff
@@ -210,11 +210,20 @@ BOOL WINAPI CreateProcessFixup(
...
+        createProcessInAppContext = createProcessInAppContextPtr->as_boolean().get();
+    }
+
 #if _DEBUG
     Log("\tPossible injection to process %ls %d.\n", exePath.data(), processInformation->dwProcessId);
 #endif
     if (((exePath.length() >= packagePath.length()) && (exePath.substr(0, packagePath.length()) == packagePath)) ||
-        ((exePath.length() >= finalPackagePath.length()) && (exePath.substr(0, finalPackagePath.length()) == finalPackagePath)))
+        ((exePath.length() >= finalPackagePath.length()) && (exePath.substr(0, finalPackagePath.length()) == finalPackagePath)) ||
+        (createProcessInAppContext))
```
rahul89dadhich, line 226: Can we add comment here adding condition to inject runtime dll to enable all further created process to run in package context ?

rahul89dadhich (changes requested):
See if can improve the readability of code and document?

Also are not sending the attribute list in createProcess detour? to pass the same behavior to all the difference create Process call from other processes?

PsfLauncher/main.cpp
```diff
@@ -146,7 +145,22 @@ int launcher_main(PCWSTR args, int cmdShow) noexcept try
...
+        bool createProcessesInAppContext = false;
+        auto createProcessesInAppContextPtr = appConfig->try_get("InPackageContext");
+        ProcThreadAttributeList m_AttributeList;
+        if (createProcessesInAppContextPtr)
+        {
+            createProcessesInAppContext = createProcessesInAppContextPtr->as_boolean().get();
+        }
+        if (createProcessesInAppContext)
+        {
+            hr = StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE, m_AttributeList.get());
```
ravishroshanms, line 158: how is m_AttributeList setting the BreakAway flag?

PsfLauncher/Readme.md
```diff
@@ -332,6 +335,13 @@ The PSF Launcher supports the use of two special purpose "pseudo-variables". The
...
 | %MsixWritablePackageRoot% | The package specific redirection location for this user when the FileRedirectionFixup is in use. | 
 
+### InPackageContext
+This configuration allows all processes created when the executable is run to operate in the same package context. This includes any child processes that might not be part of the package. This might be useful in situations where applications span processes that are outside package but needs to be run in the same package context, or when PSF fixups need to be applied to processes external to the package.
+
+When an external process is launched by an executable, all of its app data is stored in its native app data folder and hence not containerized. This feature allows the app data of the external process to be virtualized in per user per app data folder of the package.
+
+This configuration is not needed when process spanned by the executable are already part of the current or some other package. This feature can be enabled when there is a need of running external process in the current package context.
```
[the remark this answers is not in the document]
npuvvada: You're correct. The context of some other package is not required here.

PsfLauncher/main.cpp
npuvvada: Through its constructor

npuvvada:
> See if can improve the readability of code and document?
A:  Review comments are incorporated to increase code and document readability. Also, I have added an example as suggested by you in readme to illustrate this feature. 
> Also are not sending the attribute list in createProcess detour? to pass the same behavior to all the difference create Process call from other processes?
A: No, we are including prevent break away attribute when launching psfLauncher which then applies to every process in the process tree.

rahul89dadhich (approved):

PsfLauncher/main.cpp
```diff
@@ -146,7 +145,22 @@ int launcher_main(PCWSTR args, int cmdShow) noexcept try
         LogString("Working Directory: ", currentDirectory.c_str());
 
         //THROW_IF_FAILED(StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE));
-        HRESULT hr = StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE);
+        HRESULT hr = ERROR_SUCCESS;
```
dhoehna, line 148: This is a WIN32 error code.  Please change to S_OK.

PsfLauncher/main.cpp
```diff
@@ -146,7 +145,22 @@ int launcher_main(PCWSTR args, int cmdShow) noexcept try
         LogString("Working Directory: ", currentDirectory.c_str());
 
         //THROW_IF_FAILED(StartProcess(exePath.c_str(), (L"\"" + exePath.filename().native() + L"\" " + exeArgString + L" " + args).data(), (packageRoot / dirStr).c_str(), cmdShow, INFINITE));
```
dhoehna, line 147: Remove comment.

ravishroshanms (approved):

PR #222, opened 2022-12-30
CreateProcess Argument redirection feature (#220)

npuvvada:
When an external (to the container) process is launched with arguments including appData path, these are not virtualized and in turn application tries to access files in native app data path instead of per user per app data path because of which it fails to find it.

## What changed?
This feature handles redirecting createProcess argument calls from native app data to per user per app data when corresponding files are found in per user per appdata file.

## How was the change tested?
-A test application is added which triggers cmd.exe(C:\Windows\System32\cmd.exe) to rename an existing file in appdata
folder (passed as a parameter). Renaming would be successful when Argument is redirected to per user per appdata folder.


rahul89dadhich (approved):
merging from develop

PR #226, opened 2023-01-02
Powershell startScript/EndScript  error handling

npuvvada:
When startScript or EndScript mentioned in config.json fails to run either due to execution policy for powershell not set correctly or due to any other reason, there is no warning message displayed to user and user would not know whether the mentioned script is run successfully or not.

## What changed?
A warning popup message is displayed to user when startScript/EndScript fails to run mentioning the exitCode. If the exitCode matches with execution policy related failure, a detailed warning message is displayed mentioning the error and how user could reset execution policy

## How was the change tested?
It was manually tested loading startScript in an application and confirming that a popup warning is displayed when powershell execution policy is set to Restricted.

### DIFF
--- a/PsfLauncher/StartProcessHelper.h
+++ b/PsfLauncher/StartProcessHelper.h
@@ -48,15 +48,14 @@ HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR curren
         "ERROR: Failed to create a process for %ws",
         applicationName);
 
-    if (exitCode != nullptr)
-    {
-        WaitForSingleObject(processInfo.hProcess, INFINITE);
-        GetExitCodeProcess(processInfo.hProcess, exitCode);
-    }
 
     RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE), processInfo.hProcess == INVALID_HANDLE_VALUE);
     DWORD waitResult = ::WaitForSingleObject(processInfo.hProcess, timeout);
     RETURN_LAST_ERROR_IF_MSG(waitResult != WAIT_OBJECT_0, "Waiting operation failed unexpectedly.");
+    if (exitCode != nullptr)
+    {
+        GetExitCodeProcess(processInfo.hProcess, exitCode);
+    }
     CloseHandle(processInfo.hProcess);
     CloseHandle(processInfo.hThread);
 
--- a/PsfLauncher/StartProcessHelper.h
+++ b/PsfLauncher/StartProcessHelper.h
@@ -4,7 +4,7 @@
 #include "Globals.h"
 #include <wil\resource.h>
 
-HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR currentDirectory, int cmdShow, DWORD timeout, LPPROC_THREAD_ATTRIBUTE_LIST attributeList = nullptr)
+HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR currentDirectory, int cmdShow, DWORD timeout, LPPROC_THREAD_ATTRIBUTE_LIST attributeList = nullptr, DWORD *exitCode = nullptr)
 {
 
     STARTUPINFOEXW startupInfoEx =
@@ -25,7 +25,6 @@ HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR curren
         , static_cast<WORD>(cmdShow) // wShowWindow
         }
     };
-
     PROCESS_INFORMATION processInfo{};
 
     startupInfoEx.lpAttributeList = attributeList;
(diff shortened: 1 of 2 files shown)


PsfLauncher/StartProcessHelper.h
```diff
@@ -4,7 +4,7 @@
 #include "Globals.h"
 #include <wil\resource.h>
 
-HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR currentDirectory, int cmdShow, DWORD timeout, LPPROC_THREAD_ATTRIBUTE_LIST attributeList = nullptr)
+HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR currentDirectory, int cmdShow, DWORD timeout, LPPROC_THREAD_ATTRIBUTE_LIST attributeList = nullptr, DWORD *exitCode = nullptr)
```
dhoehna, line 7: a pointer to a DWORD indicates it is both input and output.  I only see it being used on line 54.  Is it used as input anywhere?

PsfLauncher/StartProcessHelper.h
```diff
@@ -49,6 +48,12 @@ HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR curren
         "ERROR: Failed to create a process for %ws",
         applicationName);
 
+    if (exitCode != nullptr)
```
dhoehna, line 51: The process executes only if exitCode isn't null?  I can pass in any number as long as it isn't null?  How does passing in a number tell the dveloper that "This process will run as long as you pass in any number.?"

PsfLauncher/StartProcessHelper.h
npuvvada: I am passing a DWORD pointer here to collect the exitCode on launching the powershell process (only to collect the output) and use it to display the error code in PsfPowershellScriptRunner.h

PsfLauncher/StartProcessHelper.h
npuvvada: This purpose of this condition (exitCode != nullptr) is to collect the exitCode only when caller needs the exitCode and passes in a valid pointer. passing in a nullptr to GetExitCodeProcess API would result in error.

"This process will run as long as you pass in any number.?" I didn't get how its related to running our process. CreateProcess is still done irrespective of this condition. Could you please elaborate?

PsfLauncher/StartProcessHelper.h
```diff
@@ -49,9 +48,14 @@ HRESULT StartProcess(LPCWSTR applicationName, LPWSTR commandLine, LPCWSTR curren
         "ERROR: Failed to create a process for %ws",
         applicationName);
 
+
     RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE), processInfo.hProcess == INVALID_HANDLE_VALUE);
     DWORD waitResult = ::WaitForSingleObject(processInfo.hProcess, timeout);
     RETURN_LAST_ERROR_IF_MSG(waitResult != WAIT_OBJECT_0, "Waiting operation failed unexpectedly.");
+    if (exitCode != nullptr)
+    {
+        GetExitCodeProcess(processInfo.hProcess, exitCode);
```
ravishroshanms, line 57: can we get the exit code from CreateProcessW ?

npuvvada: No, createProcessW only returns 0 when process creation fails.

As per the official documentation of createProcessW:
Return value
If the function succeeds, the return value is nonzero.

If the function fails, the return value is zero. To get extended error information, call [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror).

Note that the function returns before the process has finished initialization. If a required DLL cannot be located or fails to initialize, the process is terminated. To get the termination status of a process, call [GetExitCodeProcess](https://learn.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess).

ravishroshanms: We are using RETURN_LAST_ERROR_IF_MSG macro to call CreateProcessW. Can't we use this?

npuvvada: I see this is an internal #define (\include\wil\result_macros.h) to check given condition and return from function displaying error message. Errorcode looks not captured here


#define RETURN_LAST_ERROR_IF_MSG(condition, fmt, ...)           do { if (wil::verify_bool(condition)) { __RETURN_GLE_MSG_FAIL(#condition, fmt, __VA_ARGS__); }} while ((void)0, 0)

ravishroshanms: Got it. Thanks

PsfLauncher/StartProcessHelper.h
dhoehna: If this is only for output please add _Out_ to signify this is an output parameter.

PR #228, opened 2023-01-24
Powershell first launch fix

npuvvada:
When "runOnce" is set in config, powershell runs only on first launch of the application. But, if powershell fails to run due to any reason on first launch (one reason could be powershell policy not set on device correctly), script would not be run on the subsequent launch (although the issue is identified and fixed). This would force the users to re-install the application although it need not be required. (Unless in the cases when powershell script itself has to be modified to fix the issue)

## What changed?
-firstRun status of Powershell is reset until the powershell script runs successfully once. 
The registry key 'PSFScriptHasRun' which is used to capture first run of powerShell is deleted in application package hive till powerShell script runs successfully.
-Documentation of 'runOnce' is updated.

## How was the change tested?
It was manually tested by setting powershell execution policy on device to restricted and making sure powershell script runs even on subsequent launches of the application till execution policy issue is set to RemoteSigned/Bypass

### DIFF
--- a/PsfLauncher/PsfPowershellScriptRunner.h
+++ b/PsfLauncher/PsfPowershellScriptRunner.h
@@ -198,15 +198,16 @@ private:
 			std::thread pwrShellThread = std::thread(StartProcess, nullptr, script.commandString.data(), script.currentDirectory.c_str(), script.showWindowAction, script.timeout, m_AttributeList.get(), &exitCode);
 			pwrShellThread.detach();
 		}
+
 		DWORD execPolicyFailExitCode = 0x01;
 		if (exitCode == execPolicyFailExitCode)
 		{
-			MessageBoxEx(NULL, (script.scriptPath + std::wstring(L" execution faildue to an execution policy restriction. To run the script, you may need to change the execution policy. For more information, see about_Execution_Policies at https://go.microsoft.com/fwlink/?LinkID=135170.  ")).c_str(), L"Warning", MB_OK | MB_ICONWARNING, 0);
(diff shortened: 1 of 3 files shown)


PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -194,9 +195,26 @@ class PsfPowershellScriptRunner
...
+		if (exitCode != ERROR_SUCCESS)
+		{
+			// when powershell fails to run, reset first run of the powershell(when "runOnce" is set) till powershell successfully runs once
+			THROW_IF_FAILED(resetFirstRunStatus(script.shouldRunOnce));
+		}
+
+		DWORD execPolicyFailExitCode = 0x01;
+		if (exitCode == execPolicyFailExitCode)
+		{
+			MessageBoxEx(NULL, (script.scriptPath + std::wstring(L" failed due to an execution policy restriction. To change the executing policy, execute \"Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope LocalMachine\" and re-run the application.")).c_str(), L"Package Support Framework", MB_OK | MB_ICONWARNING, 0);
```
rahul89dadhich, line 211: how about adding resetFirstRunStatus in execPolicy case

rahul89dadhich (changes requested):
check comment

PsfLauncher/PsfPowershellScriptRunner.h
npuvvada: Even in case of script errors, intention is to re-run the script on every application launch till it runs successfully

rahul89dadhich (approved):

npuvvada:
> check comment

A: Resolved

PR #231, opened 2023-03-21
Error messages for config.json and StartingScriptWrapper.ps1

snigdha01091997:
When PSF is applied to an MSIX application and if config.json is not added, proper error popup is not shown to user which might cause confusion to user in figuring out why the application fails to launch.

When user tries to execute power shell script through PSF, it is mandatory to include 'StartingScriptWrapper.ps1' as it triggers the actual user power script file. But when user missed to include this file, no appropriate error message is displayed to user which makes it difficult to find the root cause of his power shell script failing to execute. Hence, showing proper error message will solve this problem.
## What changed? 
Added error messages for missing config.json and StartingScriptWrapper.ps1
## How was the change tested?
Manually tested in local by making sure popup messages are properly displayed to users when the above files are missing.

### DIFF
--- a/PsfLauncher/PsfPowershellScriptRunner.h
+++ b/PsfLauncher/PsfPowershellScriptRunner.h
@@ -269,6 +269,10 @@ private:
 					SSWrapper = file.path();
 				}
 			}
+			if (!std::filesystem::exists(SSWrapper))
+			{
+				MessageBoxEx(NULL, L"'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. Kindly add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONWARNING, 0);
+			}
 		}
 		std::wstring commandString = L"Powershell.exe ";
 		commandString.append(scriptExecutionMode);
--- a/PsfLauncher/PsfPowershellScriptRunner.h
+++ b/PsfLauncher/PsfPowershellScriptRunner.h
@@ -271,7 +271,7 @@ private:
 			}
 			if (!std::filesystem::exists(SSWrapper))
 			{
-				MessageBoxEx(NULL, L"'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. Kindly add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONWARNING, 0);
+				MessageBoxEx(NULL, L"'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. We recommend you to add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONWARNING, 0);
 			}
 		}
 		std::wstring commandString = L"Powershell.exe ";
--- a/PsfLauncher/PsfPowershellScriptRunner.h
+++ b/PsfLauncher/PsfPowershellScriptRunner.h
@@ -269,9 +269,10 @@ private:
 					SSWrapper = file.path();
 				}
 			}
+
 			if (!std::filesystem::exists(SSWrapper))
-			{
-				MessageBoxEx(NULL, L"'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. We recommend you to add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONWARNING, 0);
+			{				
+				MessageBoxEx(NULL, L"'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. The app might continue to run but not behave as expected. Kindly add it inside the package and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONWARNING, 0);
 			}
 		}
 		std::wstring commandString = L"Powershell.exe ";
--- a/PsfRuntime/Config.cpp
+++ b/PsfRuntime/Config.cpp
@@ -299,6 +299,12 @@ void load_json()
                 }
             }
         }
+        if (!file)
+        {
+            MessageBoxEx(NULL, L"‘Config.json’ was not found. Kindly add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONERROR, 0);
+            psf::TraceLogExceptions("PSFConfigException", "config.json was not found");
+            return;
+        }
     }
 
 
--- a/PsfRuntime/Config.cpp
+++ b/PsfRuntime/Config.cpp
@@ -303,6 +303,7 @@ void load_json()
         {
             MessageBoxEx(NULL, L"'Config.json' was not found. We recommend you to add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONERROR, 0);
             psf::TraceLogExceptions("PSFConfigException", "config.json was not found");
+            Log("config.json was not found");
             return;
         }
     }
--- a/PsfRuntime/Config.cpp
+++ b/PsfRuntime/Config.cpp
@@ -299,12 +299,12 @@ void load_json()
                 }
             }
         }
+
         if (!file)
         {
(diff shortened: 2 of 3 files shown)


PsfRuntime/Config.cpp
```diff
@@ -299,6 +299,12 @@ void load_json()
                 }
             }
         }
+        if (!file)
+        {
+            MessageBoxEx(NULL, L"�Config.json� was not found. Kindly add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONERROR, 0);
```
rahul89dadhich, line 304: Can we removed this unsupported chars from message?

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -269,6 +269,10 @@ class PsfPowershellScriptRunner
 					SSWrapper = file.path();
 				}
 			}
+			if (!std::filesystem::exists(SSWrapper))
+			{
+				MessageBoxEx(NULL, L"'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. Kindly add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONWARNING, 0);
```
rahul89dadhich, line 274: is StartingScriptWrapper.ps1 file needs to be in root folder?

rahul89dadhich (changes requested):
Can you please check the comment?

PsfRuntime/Config.cpp
snigdha01091997: Done the change.

PsfLauncher/PsfPowershellScriptRunner.h
snigdha01091997: Changed the message accordingly.

rahul89dadhich (approved):
looks good

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -269,6 +269,10 @@ class PsfPowershellScriptRunner
 					SSWrapper = file.path();
 				}
 			}
+			if (!std::filesystem::exists(SSWrapper))
```
dhoehna, line 272: Put a new line after the closing brace.

PsfLauncher/PsfPowershellScriptRunner.h
```diff
@@ -269,6 +269,10 @@ class PsfPowershellScriptRunner
 					SSWrapper = file.path();
 				}
 			}
+			if (!std::filesystem::exists(SSWrapper))
+			{
+				MessageBoxEx(NULL, L"'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. We recommend you to add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONWARNING, 0);
```
dhoehna, line 274: MessageBoxEx does not stop execution.  PSF will continue to execute even if starting script wrapper can't be found.

What is the expected behavior if the script can't be found?  Stop PSF, issue a warning but keep going?

I recommend to stop execution.  Apps need this script to run before start up and apps expect the script to be ran.  PSF should halt if it can't run the powershell script.

PsfRuntime/Config.cpp
```diff
@@ -299,6 +299,13 @@ void load_json()
                 }
             }
         }
+        if (!file)
+        {
+            MessageBoxEx(NULL, L"'Config.json' was not found. We recommend you to add it under the package root folder and re-install the application.", L"MSIX Packet Support Framework", MB_OK | MB_ICONERROR, 0);
```
dhoehna, line 304: Same thing here.  Once the user clicks out of the message box PSF execution will continue and PSF will crash.

PsfRuntime/Config.cpp
```diff
@@ -299,6 +299,13 @@ void load_json()
                 }
             }
         }
+        if (!file)
```
dhoehna, line 302: Add a newline after the closing brace.

PsfLauncher/PsfPowershellScriptRunner.h
dhoehna, line 274: NIT: instead of saying "Package root" put the file path, if possible, or, at least, a path inside the package.  No need to have the developer guess if we can tell them where to put it.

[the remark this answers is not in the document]
dhoehna: I mean.  A developer could make a new StartingScriptWrapper if they wanted to.   However, modifying StartingScriptWrapper.ps1 isn't something a developer needs to do for PSF.

PsfLauncher/PsfPowershellScriptRunner.h
dhoehna, line 272: NIT: Move this check up to Initialize.  Sooner we check for the existence of a required file, the better.

dhoehna (changes requested):
PSF should stop executing if config.json or StartingScriptWrapper.ps1 can't be found.

PsfLauncher/PsfPowershellScriptRunner.h
snigdha01091997: Moving this to intiailize would require additional evaluation for wrapperScript path while executing the powershell script(or adding one more class attribute for storing wrapper script path). To avoid re-duplication we remained it this way. Let me know your opinion.

PsfLauncher/PsfPowershellScriptRunner.h
snigdha01091997: In many case, even through the PowerShell script fails, it is observed that we can go and start the actual application. Hence we are not making it mandatory to halt PSF but continuing to execute the actual application by giving a warning.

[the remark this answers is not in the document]
snigdha01091997: After discussion with PMs, the following display messages are suggested: 

'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. Kindly add it under the package by downloading it again from the master release of PSF and then re-install the application. 

Let me know you opinion on this.

snigdha01091997: After discussion with PMs, the following display messages are suggested: 

**For StartingScriptWrapper.ps1:** 
'StartingScriptWrapper.ps1' was not found. This is required to run the PowerShell scripts through PSF. Kindly add it under the package by downloading it again from the master release of PSF and then re-install the application. 

**And for Config.json:**
‘Config.json’ was not found. Kindly add it in the package and re-install the application.

Let me know you opinion on this.

PsfLauncher/PsfPowershellScriptRunner.h
dhoehna: Ic ic.  Looking at it, I can see that moving this would be a bit hard.  Keep it here, but moving it up in the future would be awesome.

PsfLauncher/PsfPowershellScriptRunner.h
dhoehna: Correct.  I'm worried about state not being setup and the app either behaving differently.

Can you please add something like "The app will continue to run but the app might not behave as expected."?  To let the user know what might happen?

dhoehna: > Kindly add it in the package and re-install the application.

Can config.json be added anywhere, or in a specific spot?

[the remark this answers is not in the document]
dhoehna: Comments added above.

rahul89dadhich: As per our understanding PSF looks for config file in full package, hence we added comment "'Config.json' was not found. Kindly add it in package and re-install the application."

https://github.com/microsoft/MSIX-PackageSupportFramework/blob/master/PsfRuntime/Config.cpp#L265

PsfLauncher/PsfPowershellScriptRunner.h
snigdha01091997: Done the change

PsfRuntime/Config.cpp
snigdha01091997: Done the change

PR #243, opened 2023-10-12
Framework Package Support

SaiiDey:
Implementing framework package lookup in PSF. Apps looking for other apps via registry or environment variables can work with this. You still need to know how the main application tries to look for dependency data and add them to config


fixups/RegLegacyFixups/dllmain.cpp
```diff
@@ -19,12 +18,35 @@ bool trace_function_entry = false;
 bool m_inhibitOutput = false;
 bool m_shouldLog = true;
 
-void InitializeFixups();
```
kyelamarthi1, line 22: I am assuming the log removed here is not being used for any BI dashboards internally - Please confirm on the same.

tests/common/dependency_helper.h
```diff
@@ -0,0 +1,82 @@
+#pragma once
+
+#include <optional>
+
+#include <winrt/base.h>
+#include <winrt/Windows.Management.Deployment.h>
+#include <winrt/Windows.ApplicationModel.h>
+#include <winrt/Windows.Foundation.Collections.h>
```
kyelamarthi1, line 8: Looks like these files need to be compiled only once - Is there a possibility to add something similar to pch.h?

kyelamarthi1 (approved):
LGTM

fixups/RegLegacyFixups/dllmain.cpp
SaiiDey: Yes the logs are not related to BI dashboards at all

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -163,23 +163,53 @@ void InitializeConfiguration()
...
+                    }
+                    else if (!registry && !dependency)
+                    {
+                        Log("EnvVarFixup: Bad config. Neither 'useregistry' and 'dependency' specified");
+                        continue;
+                    }
+                    else if (registry)
                     {
-                        g_envvar_envVarSpecs.back().useregistry = true;
+                        auto useregistry = registry->as_string().wstring();
```
dhoehna, line 181: Please use the correct casing `userRegistry`

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -163,23 +163,53 @@ void InitializeConfiguration()
...
+                        traceDataStream << " useregistry: " << useregistry << " ;";
+                        LogString(0, "GetEnvFixup Config: name", variablenamePattern.data());
+                        LogString(0, "GetEnvFixup Config: value", variablevalue.data());
+                        LogString(0, "GetEnvFixup Config: useregistry", useregistry.data());
+                        g_envvar_envVarSpecs.emplace_back();
+                        g_envvar_envVarSpecs.back().remediationType = Env_Remediation_Type_Registry;
+                        g_envvar_envVarSpecs.back().variablename.assign(variablenamePattern.data(), variablenamePattern.length());
+                        g_envvar_envVarSpecs.back().variablevalue = variablevalue;
+                        if (useregistry.compare(L"true") == 0 ||
```
dhoehna, line 190: Instead of comparing against every permutation of `true` use a case insensitive compare.

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -163,23 +163,53 @@ void InitializeConfiguration()
...
+                    {
+                        Log("EnvVarFixup: Bad config. Neither 'useregistry' and 'dependency' specified");
+                        continue;
+                    }
+                    else if (registry)
                     {
-                        g_envvar_envVarSpecs.back().useregistry = true;
+                        auto useregistry = registry->as_string().wstring();
+                        traceDataStream << " useregistry: " << useregistry << " ;";
+                        LogString(0, "GetEnvFixup Config: name", variablenamePattern.data());
```
dhoehna, line 183: Please use `.c_str()` to get a pointer to the string. `.data()` does not guarantee a null terminator.

dhoehna, line 183: Please use correct casing `variableNamePattern`

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -163,23 +163,53 @@ void InitializeConfiguration()
...
+                        LogString(0, "GetEnvFixup Config: value", variablevalue.data());
```
dhoehna, line 184: Please use correct casing `variableValue`

fixups/EnvVarFixup/InitializeFixup.cpp
dhoehna, line 190: On second though, this comparison is odd.  Looks like `userRegustry` is a string when, in reality, it holds a bool.  Please change `userRegistry` to a bool.

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -163,23 +163,53 @@ void InitializeConfiguration()
...
+                        LogString(0, "GetEnvFixup Config: useregistry", useregistry.data());
+                        g_envvar_envVarSpecs.emplace_back();
+                        g_envvar_envVarSpecs.back().remediationType = Env_Remediation_Type_Registry;
+                        g_envvar_envVarSpecs.back().variablename.assign(variablenamePattern.data(), variablenamePattern.length());
+                        g_envvar_envVarSpecs.back().variablevalue = variablevalue;
+                        if (useregistry.compare(L"true") == 0 ||
+                            useregistry.compare(L"True") == 0 ||
+                            useregistry.compare(L"TRUE") == 0)
+                        {
+                            g_envvar_envVarSpecs.back().useregistry = true;
```
dhoehna, line 194: I'm confused on what the meaning of `false` means.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,102 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
+
+    Package pkg = Package::Current();
+    IVectorView<Package> dependencies = pkg.Dependencies();
+
+    std::wregex dependency_version_regex = std::wregex(L"%dependency_version%");
+    std::wregex dependency_path_regex = std::wregex(L"%dependency_root_path%");
+
+    Package dependency = nullptr;
+
+#ifdef _DEBUG
```
dhoehna, line 143: Please consolodate all `#if`'s to one method.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,102 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
+            std::wstring value_data = std::regex_replace(it.second, dependency_path_regex, dependency_path);
+            value_data = std::regex_replace(value_data, dependency_version_regex, dependency_version);
+
+            st = ::RegSetValueExW(res, it.first.c_str(), 0, REG_SZ, (LPBYTE)value_data.c_str(),
+                (DWORD)(value_data.size() + 1 /* for null termination char */) * sizeof(wchar_t));
+
+            if (st != ERROR_SUCCESS)
+            {
+#ifdef _DEBUG
+                Log("set value %s failed\n", it.first.c_str());
```
ravishroshanms, line 197: Even here we should add Trace exception logs?

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,102 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
+
+    std::wregex dependency_version_regex = std::wregex(L"%dependency_version%");
+    std::wregex dependency_path_regex = std::wregex(L"%dependency_root_path%");
+
+    Package dependency = nullptr;
+
+#ifdef _DEBUG
+    Log("Filtering required dependency\n");
+#endif
+    for (auto&& dep : dependencies) {
```
dhoehna, line 146: Please place the brace on its own line.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,102 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
+
+        if (name.find(dependency_name) != std::wstring::npos) {
+#ifdef _DEBUG
+            Log("Found required dependency\n");
+#endif
+            dependency = dep;
+            break;
+        }
+    }
+    if (dependency == nullptr)
```
dhoehna, line 160: If `dependency` is a pointer please remove `== nullptr`

tests/scenarios/MiddlewareTestA/config.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ravishroshanms, line 1: We can remove this file.

tests/scenarios/MiddlewareTestW/config.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ravishroshanms, line 1: This can be deleted too.

tests/tests.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29503.13
+# Visual Studio Version 17
+VisualStudioVersion = 17.7.34221.43
```
ravishroshanms, line 4: Why did we change this?

ravishroshanms (changes requested):

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,102 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
+
+        g_regRedirectedHivePaths.insert(narrow(reg_path));
+        g_regCreatedKeysOrdered.push_back(reg_path);
+
+        for (const auto& it : reg_entry.values)
+        {
+            std::wstring value_data = std::regex_replace(it.second, dependency_path_regex, dependency_path);
+            value_data = std::regex_replace(value_data, dependency_version_regex, dependency_version);
+
+            st = ::RegSetValueExW(res, it.first.c_str(), 0, REG_SZ, (LPBYTE)value_data.c_str(),
```
dhoehna, line 191: Do not use C-style casts.  Please use C++ casting.  Refer to here: https://github.com/microsoft/WindowsAppSDK/blob/main/dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.cpp#L235 for an example.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,102 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
+}
+
+void CleanupFixups()
+{
+    Log("RegLegacyFixups CleanupFixups()\n");
+
+    // Deleted the created keys in reverse order
+    std::vector<std::wstring>::reverse_iterator redirected_path = g_regCreatedKeysOrdered.rbegin();
+
+    while (redirected_path != g_regCreatedKeysOrdered.rend())
```
dhoehna, line 211: This can be used in a foreach loop, ya?  That way you don't need to worry about the iterator position.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -255,7 +349,35 @@ void InitializeConfiguration()
 
                             specItem.remediationRecords.push_back(recordItem);
                         }
+                        else if (type.compare(L"Redirect") == 0)
```
dhoehna, line 352: Please use a case-insensitive compare.

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -257,6 +261,28 @@ REGSAM RegFixupSam(std::string keypath, REGSAM samDesired, DWORD RegLocalInstanc
...
+
+    std::string normalizedKeypath = ReplaceRegistrySyntax(keypath);
+    size_t subkeyOffset = normalizedKeypath.find_first_of('\\');
+    if (subkeyOffset != std::wstring_view::npos)
+    {
+        std::string requestedSubkey = normalizedKeypath.substr(subkeyOffset + 1);
+        auto it = g_regRedirectedHivePaths.find(requestedSubkey);
+        if (it != g_regRedirectedHivePaths.end())
+        {
+            *key = HKEY_CURRENT_USER;
```
dhoehna, line 277: Why modify key to HKCU?

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -257,6 +261,28 @@ REGSAM RegFixupSam(std::string keypath, REGSAM samDesired, DWORD RegLocalInstanc
     return samModified;
 }
 
+template <typename CharT>
+const char* ReplaceRegistryQueryPath(PHKEY key, const CharT* subKey)
+{
+    std::string keypath = InterpretKeyPath(*key) + "\\" + InterpretStringA(subKey);
```
dhoehna, line 267: crash if key is `nullptr`

dhoehna, line 267: crash if `subkey` is null

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -257,6 +261,28 @@ REGSAM RegFixupSam(std::string keypath, REGSAM samDesired, DWORD RegLocalInstanc
...
+    std::string normalizedKeypath = ReplaceRegistrySyntax(keypath);
+    size_t subkeyOffset = normalizedKeypath.find_first_of('\\');
+    if (subkeyOffset != std::wstring_view::npos)
+    {
+        std::string requestedSubkey = normalizedKeypath.substr(subkeyOffset + 1);
+        auto it = g_regRedirectedHivePaths.find(requestedSubkey);
+        if (it != g_regRedirectedHivePaths.end())
+        {
+            *key = HKEY_CURRENT_USER;
+            return it->c_str();
```
dhoehna, line 278: I'm iffy on returning a pointer.  If the original string is deleted the returned pointer will be invalid.

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -647,7 +677,14 @@ LSTATUS __stdcall RegOpenKeyTransactedFixup(
     Log("[%d] RegOpenKeyTransacted:\n", RegLocalInstance);
 #endif
 
+    const char* updatedSubKey = ReplaceRegistryQueryPath(&key, subKey);
+    if (updatedSubKey)
```
dhoehna, line 681: Looks like line 596 and 680 are the only places where `ReplaceRegistryQueryPath` is called and, in both instances `key` is not used.  Maybe change the arguments of `ReplaceRegistryQueryPath to HKEY instead of PHKEY.`

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -1091,6 +1128,22 @@ LSTATUS __stdcall  RegGetValueFixup(
 #ifdef _DEBUG
         Log("[%d] RegGetValue:\n", RegLocalInstance);
 #endif
+
+        // If subkey is in redirected path, use it to get Registry Value
+		const char* updatedSubKey = ReplaceRegistryQueryPath(&key, SubKey);
```
dhoehna, line 1133: This code block is super tabbed.  un-tab to match indentation.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,102 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
+{
+    Log("RegLegacyFixups CleanupFixups()\n");
+
+    // Deleted the created keys in reverse order
+    std::vector<std::wstring>::reverse_iterator redirected_path = g_regCreatedKeysOrdered.rbegin();
+
+    while (redirected_path != g_regCreatedKeysOrdered.rend())
+    {
+        LSTATUS st = ::RegDeleteTreeW(HKEY_CURRENT_USER, redirected_path->c_str());
+        if (st != ERROR_SUCCESS)
```
dhoehna, line 214: Please use Windows Implementation Library for testing against error values.

tests/common/dependency_helper.h
```diff
@@ -0,0 +1,82 @@
...
+
+struct PackageModel
+{
+    std::wstring pkg_name;
+    std::wstring pkg_version;
+    std::wstring pkg_publisher;
+    std::wstring pkg_install_path;
+};
+
+inline std::optional<PackageModel> query_package_with_name(std::wstring_view pkgFamilyName)
```
dhoehna, line 27: Do not use `_` in name.  `QueryPackageWithName`

dhoehna, line 27: `const`

tests/scenarios/MiddlewareTestA/MiddlewareA.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    test_initialize("MiddlewareW Tests", 4);
+
+    if (auto vclib_package = query_package_with_name(VCLibsName))
+    {
+        {
+            test_begin("Middleware test RegQueryValue able to fetch HKLM data created by the fixup with proper replacements");
+
+            HKEY res;
+            if (RegOpenKeyExA(HKEY_LOCAL_MACHINE, Registry_Key_1, 0, KEY_ALL_ACCESS, &res) == ERROR_SUCCESS)
```
dhoehna, line 36: Please use correct WIL MACRO's for status comparison.

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: Yes it is a string. Check the [examples](https://github.com/microsoft/MSIX-PackageSupportFramework/blob/master/fixups/EnvVarFixup/readme.md#json-examples) in README. Changing it is backward incompatible and break any existing config.json. If we need to do it, it should be a separate feature.

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: Whether you should be using registry to read and write registry env variable or use the value provided in config.json

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -163,23 +163,53 @@ void InitializeConfiguration()
                     auto variablevalue = specObject.get("value").as_string().wstring();
                     traceDataStream << " value: " << variablevalue << " ;";
 
-                    auto useregistry = specObject.get("useregistry").as_string().wstring();
```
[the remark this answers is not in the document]
SaiiDey: True. Now it will still load the dll but only ignore the one with faulty config. Does this need to be changed?

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: config.json is case sensitive, right?

fixups/RegLegacyFixups/RegistryFixups.cpp
SaiiDey: Because HKLM cannot be written to without admin permission. Don't want to force that constraint to application.

fixups/RegLegacyFixups/RegistryFixups.cpp
SaiiDey: Its a pointer to a global variable. Original string is kept throughout the lifetime of the fixup.

fixups/RegLegacyFixups/RegistryFixups.cpp
SaiiDey: Updated `key `is used in the next registry call at line 599 and 683

fixups/RegLegacyFixups/RegistryFixups.cpp
SaiiDey: How? https://github.com/microsoft/MSIX-PackageSupportFramework/blob/81180e351e600afab9eb3ffb80661d36322b6f77/fixups/RegLegacyFixups/Logging.h#L250

SaiiDey: How? We are always calling this function with the pointer to HKEY we got from the main application as argument. `key` over here will always point to the argument passed on to us and never be null.

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: Its of type std::string_view. The type itself does not guarantee null terminator.

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: I am conditionally compiling log printing. How do you consolidate it?

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: This is LSTATUS not HRESULT. Which wil API do you use to check this?

tests/scenarios/MiddlewareTestA/MiddlewareA.cpp
SaiiDey: Which one? It returns a LSTATUS not HRESULT

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: Done

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: Done

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: Done

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: Updated

tests/tests.sln
SaiiDey: It got changed automatically when adding new project to test.sln. Solution files are

fixups/EnvVarFixup/EnvVarFixup.cpp
```diff
@@ -54,156 +59,244 @@ DWORD __stdcall GetEnvironmentVariableFixup(_In_ const CharC* lpName, _Inout_ Ch
...
+                            auto version = depedency.Id().Version();
+                            std::wstring dependency_version = std::to_wstring(version.Major) + L"." + std::to_wstring(version.Minor) + L"." + std::to_wstring(version.Build) + L"." + std::to_wstring(version.Revision);
+                            Log("Dependency path: %LS, version: %LS\n", dependency_path.c_str(), dependency_version.c_str());
+
+                            std::wregex dependency_version_regex = std::wregex(L"%dependency_version%");
+                            std::wregex dependency_path_regex(L"%dependency_root_path%");
+
+                            std::wstring value_data = std::regex_replace(std::wstring(spec.variablevalue), dependency_path_regex, dependency_path);
+                            value_data = std::regex_replace(value_data, dependency_version_regex, dependency_version);
+
```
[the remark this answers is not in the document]
SaiiDey: Updated

fixups/RegLegacyFixups/dllmain.cpp
```diff
@@ -19,12 +18,35 @@ bool trace_function_entry = false;
...
+
+int __stdcall PSFUninitialize() noexcept try
+{
+    CleanupFixups();
+    psf::detach_all();
+    return ERROR_SUCCESS;
+}
+catch (...)
+{
+    return win32_from_caught_exception();
```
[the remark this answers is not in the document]
SaiiDey: Added

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: Updated

fixups/RegLegacyFixups/RegistryFixups.cpp
SaiiDey: Updated

tests/common/dependency_helper.h
SaiiDey: Updated

SaiiDey: Added

tests/scenarios/MiddlewareTestA/MiddlewareA.cpp
```diff
@@ -0,0 +1,209 @@
...
+#define Environment_Variable_1         "LIB_PATH"
+#define Environment_Variable_2         "PKG_VER"
+
+int main(int argc, const char** argv)
+{
+    auto result = parse_args(argc, argv);
+
+    test_initialize("MiddlewareW Tests", 4);
+
+    if (auto vclib_package = query_package_with_name(VCLibsName))
```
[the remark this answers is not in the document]
SaiiDey: Updated

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: Updated

tests/scenarios/MiddlewareTestA/config.xml
SaiiDey: Removed

tests/scenarios/MiddlewareTestW/config.xml
SaiiDey: Removed

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: Anyway I removed them. Log implementation is anyway behind _DEBUG

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: Added

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -13,8 +13,12 @@
 #include "Logging.h"
 #include <regex>
 #include "psf_tracelogging.h"
+#include "pch.h"
+
 
 DWORD g_RegIntceptInstance = 0;
+std::unordered_set<std::string> g_regRedirectedHivePaths;
```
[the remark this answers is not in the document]
SaiiDey: Refactored them to be a part of Reg_Remediation_Record

fixups/EnvVarFixup/InitializeFixup.cpp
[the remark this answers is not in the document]
ravishroshanms: It should be fine. @npuvvada @rahul89dadhich  what do you think?

tests/tests.sln
ravishroshanms: I think we should revert this change as the pipeline might not be able to run the tests. Hopefully new added tests would still run?

fixups/EnvVarFixup/InitializeFixup.cpp
dhoehna: All `LogStrings` are expecting a const char*.  This will crash.  Either
1. Reconsider using a string_view.
2. Convert the string_view to a string, then from there get c_str()
3. Add a null terminator check in LogString.

fixups/RegLegacyFixups/InitializeFixup.cpp
dhoehna: > I am conditionally compiling log printing. How do you consolidate it?

My suggestion would be to always call log and in `log` use `#if`

fixups/RegLegacyFixups/InitializeFixup.cpp
dhoehna: True.  Sorry, with string comparisons I always default to case-insensitive.

As long as PSF won't crash if a user puts in `redirection` with a little r I'm okay with this.

fixups/RegLegacyFixups/RegistryFixups.cpp
dhoehna: Just because you're calling this now with a non-null PHKEY does not mean, in the future this will always be called with a non-null PHKEY.

Additionally, this is OSS.  Other people can fork this and make their own and they might not know that PHKEY should be not-null.

Additionally, checking for a non-null pointer is good practice against crashing.

dhoehna: Interesting. I would've looked at `IntepretStringA`  Just because code now checks for a null pointer, does not mean it'll change and subkey is used without checking if `subKey` is null.  

Please add a comment mentioning why no null check.  Something like "No null check is made because the only place subkey is used is IntepretStringA and they handles nullptr"

fixups/RegLegacyFixups/RegistryFixups.cpp
dhoehna: I see.  Sorry.  Why does `ReplaceRegistryQueryPath` change key to HKCU?

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: Yah it won't crash but simply ignore

fixups/EnvVarFixup/EnvVarFixup.cpp
```diff
@@ -54,156 +59,229 @@ DWORD __stdcall GetEnvironmentVariableFixup(_In_ const CharC* lpName, _Inout_ Ch
...
                         }
-                        // allow to fall through to system
-                        //return result;
-                        result = 0;
+
                     }
+                    else if (spec.remediationType == Env_Remediation_Type_Dependency)
+                    {
+                        Package pkg = Package::Current();
+                        IVectorView<Package> dependencies = pkg.Dependencies();
```
npuvvada, line 218: Do we have to do this operation of querying all package dependencies and resolving dependency_path and dependency_version on every GetEnvironmentVariableFixup call. Instead if it is done in initialize_config, it can be one time operation ?

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -255,12 +358,44 @@ void InitializeConfiguration()
 
                             specItem.remediationRecords.push_back(recordItem);
                         }
+                        else if (type.compare(L"Redirect") == 0)
```
npuvvada, line 361: Update traceDataStream with this fixupConfig to capture telemetry for this fixup Type.

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -564,13 +600,15 @@ LSTATUS __stdcall RegOpenKeyExFixup(
...
     Log("[%d] RegOpenKeyEx:\n", RegLocalInstance);
+    auto [updatedSubKey, keypath] = ReplaceRegistryQueryPath(&key, subKey);
+    if (updatedSubKey)
+    {
+        return RegOpenKeyExImpl(key, updatedSubKey, options, samDesired, resultKey);
+    }
 
-
-    std::string keypath = InterpretKeyPath(key) + "\\" + InterpretStringA(subKey);
     std::string registryPath = ReplaceRegistrySyntax(keypath);
```
npuvvada, line 612: This is covered now in ReplaceRegistryQueryPath, can we avoid this duplicate call ?

fixups/RegLegacyFixups/RegistryFixups.cpp
```diff
@@ -1091,8 +1134,23 @@ LSTATUS __stdcall  RegGetValueFixup(
 #ifdef _DEBUG
         Log("[%d] RegGetValue:\n", RegLocalInstance);
 #endif
+
+        // If subkey is in redirected path, use it to get Registry Value
+        auto [updatedSubKey, keypath] = ReplaceRegistryQueryPath(&key, SubKey);
+        if (updatedSubKey)
+        {
+            if constexpr (psf::is_ansi<CharT>)
```
npuvvada, line 1142: should we similarly retain ansi/wide char calls in RegOpenKeyEx and RegOpenKeyTransactedEx detours too?

fixups/RegLegacyFixups/RegistryFixups.cpp
SaiiDey: Updated

npuvvada (changes requested):
Add relevant readme sections in Registry Legacy and Environment variable fixups

fixups/EnvVarFixup/EnvVarFixup.cpp
SaiiDey: Agreed. I did it this way because I didn't feel the extra complexity is worth the performance gain.
This implementation is not too slow. Dependencies are populated and maintained by the OS. We simple read that data and do some do some string manipulations. Also this is not done for every GetEnvironmentVariableFixup call, only those that are mentioned in config.json.
Most common scenario for this feature is to read an environment variable for initializing the app, so not too performance critical.

rahul89dadhich (approved):
few NIT picking

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: Added overloaded LogString implementation which can print string_view. No need to allocate another string with null pointer. Using that to print string_view.

fixups/EnvVarFixup/InitializeFixup.cpp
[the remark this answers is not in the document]
SaiiDey: Updated. We throw now.

tests/tests.sln
SaiiDey: Reverted

fixups/RegLegacyFixups/RegistryFixups.cpp
SaiiDey: Still not convinced we should add a a check even though all code paths guarantee non null parameter.  This is an irrecoverable error and we should crash if called with null PHKEY since we cannot read or write the underlying HKEY.

Added required annotations to the parameters so that compiler can raise warnings if it finds incorrect usage of the function. It will also indicate the assumptions made by the function explicit in the signature itself.

SaiiDey: Added _In_opt_ annotation explicitly denote that `subKey` can be null

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: Added

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -157,29 +157,59 @@ void InitializeConfiguration()
...
+                    auto dependency = specObject.try_get("dependency");
+
+                    if (registry && dependency)
+                    {
+                        Log("EnvVarFixup: Bad config. Both 'useregistry' and 'dependency' specified");
+                        continue;
+                    }
+                    else if (!registry && !dependency)
+                    {
+                        Log("EnvVarFixup: Bad config. Neither 'useregistry' and 'dependency' specified");
```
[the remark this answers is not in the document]
SaiiDey: Done

fixups/EnvVarFixup/EnvVarFixup.cpp
```diff
@@ -54,156 +59,229 @@ DWORD __stdcall GetEnvironmentVariableFixup(_In_ const CharC* lpName, _Inout_ Ch
...
-                                    RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_ZEROONFAILURE, &type, lpValue, &dLen);
-                                if (ret == ERROR_SUCCESS)
+                                if (RegOpenKeyExA(HKEY_CURRENT_USER, "Environment",
+                                    0, KEY_ENUMERATE_SUB_KEYS | KEY_READ | KEY_QUERY_VALUE, &hKeyCU) == ERROR_SUCCESS)
                                 {
-                                    Log("[%d] GetEnvironmentVariableFixup:(A) HKCU value found %s", GetEnvVarInstance, lpValue);
-                                    result = dLen;
+                                    Log("[%d] GetEnvironmentVariableFixup:(A) HKCU key found.", GetEnvVarInstance);
+                                    DWORD type = RRF_RT_REG_SZ;
+                                    DWORD dLen = (DWORD)lenBuf;
```
dhoehna, line 76: Please do not use c-style casts.

fixups/EnvVarFixup/EnvVarFixup.cpp
```diff
@@ -54,156 +59,229 @@ DWORD __stdcall GetEnvironmentVariableFixup(_In_ const CharC* lpName, _Inout_ Ch
...
+                                    auto ret = RegGetValueA(HKEY_CURRENT_USER, "Environment", lpName,
+                                        RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_ZEROONFAILURE, &type, lpValue, &dLen);
+                                    if (ret == ERROR_SUCCESS)
```
dhoehna, line 79: Please use WIL macro for testing error codes.

fixups/EnvVarFixup/InitializeFixup.cpp
```diff
@@ -157,29 +165,57 @@ void InitializeConfiguration()
...
+                        auto useRegistry = registry->as_string().wstring();
+                        traceDataStream << " useregistry: " << useRegistry << " ;";
+                        LogString(0, "GetEnvFixup Config: name", variableNamePattern);
+                        LogString(0, "GetEnvFixup Config: value", variableValue);
+                        LogString(0, "GetEnvFixup Config: useregistry", useRegistry);
+                        g_envvar_envVarSpecs.emplace_back();
+                        g_envvar_envVarSpecs.back().remediationType = Env_Remediation_Type_Registry;
+                        g_envvar_envVarSpecs.back().variablename.assign(variableNamePattern.data(), variableNamePattern.length());
+                        g_envvar_envVarSpecs.back().variablevalue = variableValue;
+                        if (useRegistry.compare(L"true") == 0 ||
```
dhoehna, line 196: If you change `userRegustry` to all lowercase you only need one comparison.

fixups/RegLegacyFixups/InitializeFixup.cpp
```diff
@@ -117,12 +121,111 @@ void LogString(const wchar_t* name, const wchar_t* value)
...
 
+struct EntryToCreate 
+{
+    std::wstring path;
+    std::unordered_map<std::wstring, std::wstring> values;
+};
+
 
-void InitializeFixups()
+Redirect_Registry CreateRegistryRedirectEntries(std::wstring_view dependency_name, std::vector<EntryToCreate> entries)
```
dhoehna, line 131: NIT use const reference.

dhoehna:
I'm concerned about adding WinRT to PSF.  One feature of PSF is allowing apps to work down level.  Is this still true with WinRT?

fixups/EnvVarFixup/EnvVarFixup.cpp
SaiiDey: Registry apis return LSTATUS not HRESULT

fixups/EnvVarFixup/InitializeFixup.cpp
SaiiDey: It will also allow `tRUe`, `TrUe`,... Do we want that?

SaiiDey:
> Add relevant readme sections in Registry Legacy and Environment variable fixups

Added. Do check them if they explanation is clear.

fixups/EnvVarFixup/EnvVarFixup.cpp
SaiiDey: Updated

fixups/RegLegacyFixups/readme.md
```diff
@@ -213,6 +250,17 @@ Here is the equivalent config section when using XML to specify.
 						<value>"^SubKey"</value>
 					</values>
 				</remediation>
+				<remediation>
+					<type>Redirect</type>
+					<dependency>java</hive>
+					<data>
+						<key>Software\\random_key\\abc\\%dependency_version%</key>
```
npuvvada, line 257: Nit: Typo error for closing tag

fixups/RegLegacyFixups/InitializeFixup.cpp
SaiiDey: My bad. updated.

fixups/RegLegacyFixups/readme.md
```diff
@@ -213,6 +250,17 @@ Here is the equivalent config section when using XML to specify.
 						<value>"^SubKey"</value>
 					</values>
 				</remediation>
+				<remediation>
+					<type>Redirect</type>
+					<dependency>java</hive>
```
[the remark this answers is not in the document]
SaiiDey: Fixed

npuvvada (approved):

rahul89dadhich (changes requested):
Wait until we get clarity on down level

rahul89dadhich (approved):

PR #251, opened 2024-01-18
Changes for migrating pipeline to 1ESPT

zinniaa:
### **Why is this change being made?**
This change is being made to migrate our pipeline from the classic pipeline to 1ES Pipeline Template. This is necessary to comply with new requirements and ensure that it is part of a governed solution. 

### **What changed?**
- Converted the tasks of the classic pipeline to YAML format, which extends the 1ES Pipeline Template. 
- Added `.gdnbaselines` file for manual baselining of the newly created 1ESPT pipeline.
- Made changes to fix violations that were causing errors in the new pipeline. 

### **How was the change tested?**
Created a new pipeline as per 1ESPT guidelines and tested it by running the build for this branch. The pipeline was verified to be running successfully and all tasks were completing without errors.

### DIFF
--- a/azure-pipelines/psf-public-nuget-build.yml
+++ b/azure-pipelines/psf-public-nuget-build.yml
@@ -1,0 +1,352 @@
+name: $(MajorVersion).$(MinorVersion).$(date:yyMMdd)$(rev:.r)
+variables:
+  - name: BuildConfiguration
+    value: release
+  - name: CertFriendlyName
+    value: CentennialFixups Test Signing Certificate
+  - name: CertPassword
+    value: CentennialFixupsTestSigning
+  - name: CertStoreLocation
+    value: Cert:\LocalMachine\My
+  - name: MajorVersion
+    value: "1"
+  - name: MinorVersion
+    value: "0"
+resources:
+  repositories:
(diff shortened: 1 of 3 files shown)


azure-pipelines/psf-public-nuget-build.yml
```diff
@@ -0,0 +1,352 @@
...
+                    *\Release\Dia2Lib.dll
+                    *\Release\Microsoft.Diagnostics.FastSerialization.dll
+                    *\Release\Microsoft.Diagnostics.Tracing.TraceEvent.dll
+                    *\Release\OSExtensions.dll
+                    *\Release\TraceReloggerLib.dll
+                  TargetFolder: $(Build.BinariesDirectory)
+                  OverWrite: true
+
+              # sign binaries
+              - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@3
```
npuvvada, line 282: ESRP codesigning task version looks to be different for sign binaries and sign nuget. Let's mantain one version for both

npuvvada (approved):
Looks good

rahul89dadhich (approved):

PR #256, opened 2024-02-16
Users/snigdhalnu/psf admin changes

snigdha01091997:
#Why were the changes made?
PSF does not work when app requires UAC prompt.
To support the applications requiring admin privileges to launch

#What changed?
Added a functionality to launch application with UAC prompt if launching them normally fails.
To use API ShellExecuteEx rather than CreateProcessW to launch application in admin mode.

#How were the changes tested?
Performed test with msix requiring admin privileges.

### DIFF
--- a/PsfLauncher/main.cpp
+++ b/PsfLauncher/main.cpp
@@ -39,6 +39,7 @@ LARGE_INTEGER psfLoad_startCounter;
 void LogApplicationAndProcessesCollection();
 int launcher_main(PCWSTR args, int cmdShow) noexcept;
 void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr);
+void LaunchInBackgroundAsAdmin(const wchar_t executable[], const wchar_t arguments[], bool wait, int cmdShow, LPCWSTR dirStr); 
 void LaunchMonitorInBackground(std::filesystem::path packageRoot, const wchar_t executable[], const wchar_t arguments[], bool wait, bool asAdmin, int cmdShow, LPCWSTR dirStr);
 bool IsCurrentOSRS2OrGreater();
 std::wstring ReplaceVariablesInString(std::wstring inputString, bool ReplaceEnvironmentVars, bool ReplacePseudoVars);
@@ -166,7 +167,11 @@ int launcher_main(PCWSTR args, int cmdShow) noexcept try
         }
         if (hr != ERROR_SUCCESS)
         {
-            Log("Error return from launching process 0x%x.", GetLastError());
+            Log("Error return from launching process second try, try again 0x%x.", GetLastError());
+        }
+        if (hr != ERROR_SUCCESS)
+        {
+            LaunchInBackgroundAsAdmin(exePath.c_str(), args, true, cmdShow, currentDirectory.c_str());
         }
     }
     else
@@ -201,6 +206,9 @@ catch (...)
     return win32_from_caught_exception();
 }
 
+
+
+
 void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path packageRoot, int cmdShow, LPCWSTR dirStr)
 {
     std::wstringstream traceDataStream;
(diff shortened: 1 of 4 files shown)


npuvvada (approved):

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,170 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+    Log("PsfLaunch: Unable to launch with injection (possibly due to elevation).  Launch using shell launch without injection.");
+    THROW_LAST_ERROR_IF_MSG(!ShellExecuteEx(&shExInfo), "Error starting monitor using ShellExecuteEx");
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE), shExInfo.hProcess == INVALID_HANDLE_VALUE);
+
+    // possibly inject the runtime now?
+    try
+    {
+        BOOL b;
+        BOOL res = IsProcessInJob(shExInfo.hProcess, nullptr, &b);
+        if (res != 0)
```
dhoehna, line 277: I highly recommend testing the error cases first.  This way you avoid using too many indents and the golden path is easier to read.  For example instead of `if (res != 0) { //continue happy path}` do `if(res == 0) {//error?  Early return?}` and for `if(b == TRUE)` do `if(b == FALSE {return}`

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,170 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+        , 0           // hInstApp
+    };
+    Log("PsfLaunch: Unable to launch with injection (possibly due to elevation).  Launch using shell launch without injection.");
+    THROW_LAST_ERROR_IF_MSG(!ShellExecuteEx(&shExInfo), "Error starting monitor using ShellExecuteEx");
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE), shExInfo.hProcess == INVALID_HANDLE_VALUE);
+
+    // possibly inject the runtime now?
+    try
+    {
+        BOOL b;
```
dhoehna, line 275: Do not use single letter variable names.

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,170 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+        if (res != 0)
+        {
+            if (b == TRUE)
+            {
+                Log("\tPsfLaunch: New process is in a job, allow injection.");
+                // Fix for issue #167: allow subprocess to be a different bitness than this process.
+                USHORT bitness = ProcessBitness(shExInfo.hProcess);
+                DWORD procId = GetProcessId(shExInfo.hProcess);
+                std::filesystem::path packageRoot = PSFQueryPackageRootPath();
+#if _DEBUG
```
dhoehna, line 286: Why `if DEBUG`? why not always log?

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,170 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+                Log("\tPsfLaunch: Injection for PID=%d Bitness=%d", procId, bitness);
+#endif  
+                std::wstring wtargetDllName = FixupDllBitness(std::wstring(psf::runtime_dll_name), bitness);
+#if _DEBUG
+                Log("\tPsfLaunch: Use runtime %ls", wtargetDllName.c_str());
+#endif
+                static const auto pathToPsfRuntime = (packageRoot / wtargetDllName.c_str()).string();
+                const char* targetDllPath = NULL;
```
dhoehna, line 294: use `nullptr` instead of `NULL`

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,170 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE), shExInfo.hProcess == INVALID_HANDLE_VALUE);
+
+    // possibly inject the runtime now?
+    try
+    {
+        BOOL b;
+        BOOL res = IsProcessInJob(shExInfo.hProcess, nullptr, &b);
+        if (res != 0)
+        {
+            if (b == TRUE)
```
dhoehna, line 279: The API does return BOOL.  However, it seems like this is a `WIN32_ERROR` behind the scenes according to the documentation.  Consider comparing b to `ERROR_SUCCESS` instead of `TRUE/FALSE`

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,170 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+                            catch (...)
+                            {
+                                Log("\tPsfLaunch: Non-fatal error enumerating directories while looking for PsfRuntime.");
+                            }
+                        }
+
+                    }
+                }
+
+                if (targetDllPath != NULL)
```
dhoehna, line 347: Please use `nullptr`

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,170 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+    if (wait)
+    {
+        WaitForSingleObject(shExInfo.hProcess, INFINITE);
+        CloseHandle(shExInfo.hProcess);
+    }
+    else
+    {
+        WaitForInputIdle(shExInfo.hProcess, 1000);
+        // Due to elevation, the process starts, relaunches, and the main process ends in under 1ms.
+        // So we'll just toss in an ugly sleep here for now.
```
dhoehna, line 388: Please remove `ugly`

PsfLauncher/main.cpp
snigdha01091997: Yes, I have done the related changes in the latest commit. Thanks

PsfLauncher/main.cpp
snigdha01091997: Rectified the variable name and renames as isProcessInJob

PsfLauncher/main.cpp
snigdha01091997: To ensure the log functions are only active during debugging, encapsulate them within a preprocessor directive like #ifdef DEBUG. This guards the logging code, activating it only when the debug flag is set, ensuring it's utilized solely for debugging purposes.

PsfLauncher/main.cpp
snigdha01091997: https://learn.microsoft.com/en-us/windows/win32/api/jobapi/nf-jobapi-isprocessinjob#parameters

According to public documentation, the out value is of BOOL type. If you have any other reference please let me know.

PsfLauncher/main.cpp
snigdha01091997: Done the changes in the latest commit.

PsfLauncher/main.cpp
snigdha01091997: Done the changes in the latest commit.

PsfLauncher/main.cpp
dhoehna: I apologize.  You interpreted my question in a different way than I was expecting it to be.  I understand what happens in code in a `#ifdef DEBUG` what I don't understand is why include `#IFDEF DEBUG` at all?  Why not have this be logged all the time?  It should help diagnose issues in development.

PsfLauncher/main.cpp
dhoehna: Yeah, you're right.  I did notice that `BOOL` was returned.  I mistakenly assumed that, because GetLastError can be used if the return value is `FALSE` that `BOOL` must have the error code inside it.  However the [documentation](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/9d81be47-232e-42cf-8f0d-7a3b29bf2eb2) says it is either a 0 or a 1.

If it really was a `WIN32` error a `DWORD` would have been returned.

My mistake.

PsfLauncher/main.cpp
snigdha01091997: Removed the debug tag in the latest commit. Thanks for the suggestion.

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,164 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
...
+    try
+    {
+        BOOL isProcessInJob;
+        BOOL res = IsProcessInJob(shExInfo.hProcess, nullptr, &isProcessInJob);
+
+        if (res == 0 || isProcessInJob == FALSE)
+        {
+            return;
+        }
+        else
```
dhoehna, line 289: `else` can be removed.  Because of the `return` statement on line 287 we know PSF is still in a good state after `return`

PsfLauncher/main.cpp
```diff
@@ -230,6 +244,164 @@ void GetAndLaunchMonitor(const psf::json_object& monitor, std::filesystem::path
     LaunchMonitorInBackground(packageRoot, monitorExecutable->as_string().wide(), monitorArguments->as_string().wide(), wait, asAdmin, cmdShow, dirStr);
 }
 
+void LaunchInBackgroundAsAdmin(const wchar_t executable[], const wchar_t arguments[], bool wait, int cmdShow, LPCWSTR dirStr)
+{
+    std::wstring cmd = L"\"";
+    cmd.append(executable);
```
dhoehna, line 250: Nothing to change, but leaving a note.  `wstring::append` expects a null-terminated string.    The code that calls `LaunchInBackgroundAsAdmin` uses null-terminated strings.  Someday in the future, this inputs should be checked for the null-termination.

